PR #262, opened 2021-05-06
.github: Fix link to CONTRIBUTING.md in bug report template

pwithnall:
The existing relative link won’t work, as it’s eventually resolved
relative to
`github.com/cloud-carbon-footprint/cloud-carbon-footprint/issues/123`
rather than to the `.github/workflows` directory in git, as I suspect
was intended.

The relative link `../blob/trunk/CONTRIBUTING.md` does work when
resolved relative to the issue URI, but is fairly unintelligible by
someone when they’re previewing an issue or reading the Markdown while
filling in the template.

So I think it makes most sense to use the full URI to `CONTRIBUTING.md`,
which will work in all situations.

Signed-off-by: Philip Withnall <pwithnall@endlessos.org>

## Description of Change

See commit message above.

## Checklist

- [x] relevant documentation is changed or added

© 2021 ThoughtWorks, Inc.

### DIFF
--- a/.github/ISSUE_TEMPLATE/Bug_report.md
+++ b/.github/ISSUE_TEMPLATE/Bug_report.md
@@ -10,7 +10,7 @@ assignees: ''
 
 <!-- Please ensure you've completed the following steps by replacing [ ] with [x]-->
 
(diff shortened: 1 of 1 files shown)


ccasher:
Nice catch! I will go ahead and merge this change. Thank you for contributing!

PR #269, opened 2021-05-10
core: Add USW2-RDS:ChargedBackupUsage to types list

pwithnall:
Signed-off-by: Philip Withnall <pwithnall@endlessos.org>

## Description of Change

Adds a missing suffix check for `RDS:ChargedBackupUsage` to set it as using a HDD. This fixes the warning:
```
api_1     | 2021-05-25T10:03:01.861Z [RDS Storage Logger] warn: Unexpected Cost explorer Dimension Name: USW2-RDS:ChargedBackupUsage 
```

## Checklist

### DIFF
--- a/packages/core/src/services/aws/RDSStorage.ts
+++ b/packages/core/src/services/aws/RDSStorage.ts
@@ -77,7 +77,11 @@ export default class RDSStorage implements ICloudService {
       awsGroupKey.endsWith('PIOPS-Storage')
     )
       return DiskType.SSD
-    if (awsGroupKey.endsWith('StorageUsage')) return DiskType.HDD
+    if (
+      awsGroupKey.endsWith('StorageUsage') ||
+      awsGroupKey.endsWith('RDS:ChargedBackupUsage')
+    )
+      return DiskType.HDD
     this.rdsStorageLogger.warn(
       'Unexpected Cost explorer Dimension Name: ' + awsGroupKey,
     )


packages/core/src/services/aws/CostAndUsageTypes.ts
```diff
@@ -53,6 +53,7 @@ export const HDD_USAGE_TYPES: string[] = [
   'GlacierByteHrs', // Glacier
   'Aurora:BackupUsage', // Aurora back up in S3
   'RDS:ChargedBackupUsage', // RDS Backup in S3
+  'USW2-RDS:ChargedBackupUsage', // RDS Backup in S3
```
dtoakley: Hi @pwithnall -- thanks for the PR and apologies for the delayed response. This usage type should be handled by the string above it: `RDS:ChargedBackupUsage`. This is because in the CostAndUsageReports class we check that the string ends with any of the options in this array: `this.endsWithAny(HDD_USAGE_TYPES, usageType)`

Let me know if I'm missing something though and it's not being handled for you!

pwithnall: I put this PR together because the server was printing a warning about `USW2-RDS:ChargedBackupUsage` not being handled, so something’s amiss. I followed the instructions in the documentation about handling those warnings to put this PR together. However, I haven’t tested it, so it could be the wrong fix.

dtoakley: Hi @pwithnall - would you be able to pull the latest from the trunk branch and re-test and see if the warning is still showing? If it is, could you also log the entire `costAndUsageReportRow`, but changing line 212 of the `CostAndUsageReport` class:

`this.costAndUsageReportsLogger.warn(`Unexpected usage type for storage service: ${costAndUsageReportRow}`)`

pwithnall: OK, I’ll give that a go when I get a chance in the next week or two, thanks.

pwithnall: I dug into it a bit more and actually chased down the warning message, and it’s not a `CostAndUsageTypes` thing at all. I’ve updated the PR with a tested fix.

dtoakley:
@pwithnall -- I made a small tweak so it's consistent with the other storage types being used, and add a test case for this usage type, and merge it in with this PR: https://github.com/cloud-carbon-footprint/cloud-carbon-footprint/pull/303

Going to close this PR and it's all looking good. Thanks so much @pwithnall -- this is a great improvement!

PR #297, opened 2021-05-26
Corrected paths for GCP and Azure definitions

davidmytton:
## Description of Change

The paths to the definition files for GCP and Azure were incorrect in the methodology docs. This corrects them.

## Checklist

- [x] relevant documentation is changed or added


dtoakley:
Thanks @davidmytton -- good catch. Merged! 

Side note: we are working on extracting the cloud providers to their own packages, so this and other references to them will need to updated again soon.

PR #358, opened 2021-06-28
fix: adds cloud provider to emissions factor result and updates unknowns

dtoakley:


packages/aws/src/lib/AWSRegions.ts
```diff
@@ -28,4 +28,5 @@ export enum AWS_REGIONS {
   SA_EAST_1 = 'sa-east-1',
   US_GOV_EAST_1 = 'us-gov-east-1',
   US_GOV_WEST_1 = 'us-gov-west-1',
+  UNKNOWN = 'unknown',
```
dtoakley, line 31: Interested in why we're adding this -- is it just in case we don't know the region in the future so it doesn't break the CO2e conversion, like is probably happening with Azure?

I think that is a good reason to add it, but I also checked the `product_region` values in the CUR data, and it includes an option for `global`, so I'm thinking we add a `global` region as well, which will likely have the same value as `unknown` as it'll be the global average.

dtoakley:
Added a comment to the `AWSRegions.ts` file, and push some improvements to the branch. However I'm not sure I understand why we need to update this endpoint with the cloud provider to help with recommendations for greener regions... All that logic will be done on the server side and returned in the response from the recommendations endpoint. We *could* do the calculation on the client side, but that would have performance implications and doesn't make all that much sense to me -- although I could be missing something so let's discuss it!

PR #491, opened 2021-09-30
fix: currentcost value wasn't formatted in forecast

jgrumboe:


4upz-tw (approved):
Thanks for catching this so fast, Johannes! I'll merge in this fix and will do a follow-up commit to apply the same formatter to the current co2e value as well.

jgrumboe:
@4upz-tw you're welcome, I just noticed that it was too fast for Hacktoberfest 😁

PR #507, opened 2021-10-02
Sorting by carbon savings when using metric tons or kg should result in the same order

### LINKED ISSUE #450: Sorting by carbon savings when using metric tons or kg should result in the same order
**Context**
In order to give a better picture of which recommendations will lead to the greatest carbon savings, we added in a feature to show it in terms of kg, which is a better fit when dealing with smaller values.
 **Description of Issue**

akhil-ghatiki:
## This pull request removes the round off the value of carbon savings and increases the precision of the value. 
Issue No : #450 
cc : @ccasher  @ericksod 

## Checklist

- [x] yarn test passes
- [x] yarn lint has been run

© 2021 Thoughtworks, Inc.


aalva05-tw:
Nice work!

PR #519, opened 2021-10-08
Adds parameters to cloudformation template

mawi:
## Description of Change

It replaces harcoded strings in the ccf-athena.yaml file with cloudformation parameters

## Checklist

## Notes

© 2021 Thoughtworks, Inc.

### DIFF
--- a/cloudformation/ccf-athena.yaml
+++ b/cloudformation/ccf-athena.yaml
@@ -1,3 +1,16 @@
+Parameters:
+  AuthorizedRole:
+    Description: The trusted entity's ARN that can assume the CCF role
+    Type: String
+    Default: arn:aws:iam::<ACCOUNT ID>:<NAME>
+  BillingDataBucket:
+    Description: The S3 bucket where the CUR data lives
+    Type: String
+    Default: arn:aws:s3:::<YOUR BUCKET NAME>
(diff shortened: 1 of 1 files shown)


cloudformation/ccf-athena.yaml
```diff
@@ -1,3 +1,16 @@
+Parameters:
+  AuthorizedRole:
+    Description: The trusted entity's ARN that can assume the CCF role
+    Type: String
+    Default: arn:aws:iam::<ACCOUNT ID>:<NAME>
```
dfratte, line 5: I think we can do something like this here

```suggestion
    Default: !Sub arn:aws:iam::${AWS::AccountId}:role/<ROLE_NAME>
```

mawi: Good tip! Didn't know that placeholder. But we can have the case where the authorized role or user doesn't live in the same account that the ccf-athena role. So that may be confusing

gasparearmato6991 (approved):

dtoakley-tw (approved):

dtoakley (approved):

dtoakley:
LGTM -- I also tested this manually as well and worked great. Merged!

PR #564, opened 2021-11-03
Add IAM auth mode

dfratte:
## Description of Change

This PR adds a new auth mode for the application. The IAM mode relies on using the permissions available to the application via an EC2 instance profile. 

It requires the application to be running on an EC2 instance that has been provisioned with the application and all env vars are setup correctly. Moreover, the EC2 instance needs to have attached a role that should have all the permissions required by the app (same role/policies as defined via CloudFormation)

## Checklist

- [x] yarn test passes
- [x] yarn lint has been run

## Notes

N/A

© 2021 Thoughtworks, Inc.

### DIFF
--- a/packages/aws/src/application/AWSCredentialsProvider.ts
+++ b/packages/aws/src/application/AWSCredentialsProvider.ts
@@ -6,6 +6,7 @@ import {
   Credentials,
   config as awsConfig,
   ChainableTemporaryCredentials,
+  EC2MetadataCredentials,
 } from 'aws-sdk'
 import { configLoader } from '@cloud-carbon-footprint/common'
 import GCPCredentials from './GCPCredentials'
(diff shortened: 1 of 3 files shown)


packages/aws/src/application/AWSCredentialsProvider.ts
```diff
@@ -30,6 +31,11 @@ export default class AWSCredentialsProvider {
               configLoader().AWS.authentication.options.targetRoleName,
           },
         })
+      case 'IAM':
```
dtoakley: Maybe `EC2-METADATA` is a better name for this credentials option? All the options have something to do with `IAM` right?

dtoakley:
Let's make sure to include an update to the documentation about these options as part of the change in `microsite/docs/AWS.md`

dfratte:
Feedback and suggestions should be addressed in last 2 commits!

ccasher (approved):
Looks great! Will test this out locally and merge to trunk.

PR #611, opened 2021-12-04
[610] Get data for azure subscriptions in series

### LINKED ISSUE #610: Azure - multiple subscriptions fails due to running in parallel
## Feature Suggestion
I am pulling in a lot of Azure subscriptions (like 50+) and the current implementation attempts to get them all in parallel, which causes lots of issues like rate limiting and out of memory.

**Describe the solution you'd like**
Re-configure it to work in series (slower, but works).

**Describe alternatives you've considered**
None

**Additional context**
None

## Possible Implementation
I will submit a PR.

cliftonc:
## Description of Change

This fixes #610 - simply runs the request for each subscription in series instead of in parallel.

## Checklist

- [X] yarn test passes
- [X] yarn lint has been run

## Notes

None

© 2021 Thoughtworks, Inc.


dtoakley:
Hey @cliftonc -- thank you so much for submitting this PR. After spending a bit of time with it, I think the major problem is that the `usageDetails.list` method in the `ConsumptionManagementService` class was hitting request limits and we didn't have any back-off and retry logic implemented like we do for the `usageDetails.listNext` method. 

I've pushed a new branch `infinitaslearning/trunk` with the retry logic in place, working off of your commit.  Initial testing with ~40 subscriptions succeeds for me when gathering all data from 2021, with all the subscriptions running in parallel, taking ~8 minutes. 

That said, if you are still having issues (e.g. memory heap), instead of running each subscription sequentially, it would be more performant to run them in groups/chunks. I have also committed that implementation in the `AzureAccount` class but commented out for now, so you can also try that out. I've also added some logging of the time for the request to process. Please re-pull that new branch and let me know how it goes for you. 

Based on your feedback and further testing, we can decide which approach to merge in and make the default for everyone :) Many thanks again!

dtoakley:
Here are the changes if you want to view them in Github too: https://github.com/cloud-carbon-footprint/cloud-carbon-footprint/pull/613/files

cliftonc:
Awesome!  I didn't quite follow the code well enough to make the change you have, so went with the simple option :)  I'll give it a go this morning (including with the chunking logic), and update the PR.

cliftonc:
@dtoakley I ran it this morning with the chunking as well, and it worked fine - it hit a couple of retries but got through.  

I don't know if its our subscriptions (?) but I get a *lot* of 'usage rows' from `getConsumptionUsageDetails` ... e.g. 95k per month for one subscription - is this normal?  The only way so far I have been able to get it to actually finish is to run it month by month manually, not for a full year.

Is there a way to get the data out that is less granular?  Or is this just how it is and I need to run it on a server instead of my laptop that goes to sleep :D

dtoakley:
Nice -- well it's good to know this is an improvement. I've reach out other Azure users to also them for feedback on the parallel vs sequential approaches, to help us decide what to settle on by default.

Re: you subscription data - yeah, the API returns a lot of results, so I'm not surprised, if you have a lot of usage in a subscription.  There isn't a way to get them with less granularity that I'm aware of at the moment. The APIs for AWS and GCP allows for this, so it's a little annoying in the context of Azure -- but we'll keep looking for better ways!

cliftonc:
Yes please do merge and close my original one.  I am struggling with the fact the hooks seem to rely on both being in trunk and talisman, and getting everything running reliably on my fork, a lot of things in this repo we don't typically use here that I need to get used to :)

dtoakley:
Great thanks @cliftonc -- sorry but I ended up adding the changes directly to trunk rather than merging, because I want that new branch to be available for others to feedback on. Next time I will make sure your commits make it to trunk so you're shown as a contributor. Closing this one for now!

cliftonc:
No problem!

PR #730, opened 2022-03-07
Updated API router to accept a CCF Config

HHogg:
## Checklist

- [ ] yarn test passes
- [ ] yarn lint has been run

## Notes

Hi 👋  As part of Spotify's Climate Change Hackweek we're looking at putting the CCF dashboard into a [Backstage plugin ](https://github.com/backstage/backstage). 

The approach we're trying out is plugging the CCF express router into Backstages express server, like so... 

```ts
import ccfRouter from '@cloud-carbon-footprint/api/dist/api';

app.use(ccfRouter({
  AZURE: {
     USE_BILLING_DATA: bsConfigLoader.get('...'),,
     authentication: {
        mode: bsConfigLoader.get('...'),
        clientId: bsConfigLoader.get('...'),
        clientSecret: bsConfigLoader.get('...'),
        tenantId: bsConfigLoader.get('...'),
      }   
  }
}));
```

Backstage has it's own configLoader system, this PR makes changes to allow a CCFConfig to be passed in and set inside the router. Please let us know your thoughts, if this is a welcomed change and/or if there's a better way to go about this. 

_The `'@cloud-carbon-footprint/api/dist/api` isn't ideal, and if this is a welcomed change, then we can look into having the router as a top level export?_

© 2021 Thoughtworks, Inc.

### DIFF
--- a/packages/api/src/api.ts
+++ b/packages/api/src/api.ts
@@ -114,7 +114,7 @@ const RecommendationsApiMiddleware = async function (
   }
 }
 
-const createRouter = (config?: CCFConfig) => {
+export const createRouter = (config?: CCFConfig) => {
   setConfig(config)
 
   const router = express.Router()
@@ -128,5 +128,3 @@ const createRouter = (config?: CCFConfig) => {
 
   return router
 }
-
-export default createRouter
--- a/packages/api/src/api.ts
+++ b/packages/api/src/api.ts
@@ -13,6 +13,8 @@ import {
 } from '@cloud-carbon-footprint/app'
 
 import {
+  setConfig,
+  CCFConfig,
   EstimationRequestValidationError,
   Logger,
   PartialDataError,
@@ -112,12 +114,19 @@ const RecommendationsApiMiddleware = async function (
   }
 }
 
-const router = express.Router()
+const createRouter = (config?: CCFConfig) => {
+  setConfig(config)
 
-router.get('/footprint', FootprintApiMiddleware)
-router.get('/regions/emissions-factors', EmissionsApiMiddleware)
-router.get('/recommendations', RecommendationsApiMiddleware)
-router.get('/healthz', (req: express.Request, res: express.Response) => {
-  res.status(200).send('OK')
-})
-export default router
+  const router = express.Router()
+
+  router.get('/footprint', FootprintApiMiddleware)
+  router.get('/regions/emissions-factors', EmissionsApiMiddleware)
+  router.get('/recommendations', RecommendationsApiMiddleware)
+  router.get('/healthz', (req: express.Request, res: express.Response) => {
+    res.status(200).send('OK')
+  })
+
+  return router
+}
+
+export default createRouter
--- a/packages/app/src/CacheManager.ts
+++ b/packages/app/src/CacheManager.ts
@@ -12,14 +12,13 @@ import EstimatorCacheGoogleCloudStorage from './EstimatorCacheGoogleCloudStorage
 
 const cacheService: EstimatorCache = new EstimatorCacheFileSystem()
 const googleCloudCacheService: EstimatorCache =
-  new EstimatorCacheGoogleCloudStorage(configLoader().GCP.CACHE_BUCKET_NAME)
+  new EstimatorCacheGoogleCloudStorage()
 
 export default class CacheManager implements EstimatorCache {
   private readonly currentCacheMode: string
   private readonly cacheLogger: Logger
 
   constructor() {
-    this.currentCacheMode = configLoader().CACHE_MODE
     this.cacheLogger = new Logger('cache')
   }
 
@@ -34,7 +33,7 @@ export default class CacheManager implements EstimatorCache {
     const { GCS } = this.cacheModes
     let estimates
 
-    switch (this.currentCacheMode) {
+    switch (configLoader().CACHE_MODE) {
       case GCS:
         this.cacheLogger.info('Using GCS bucket cache file...')
         estimates = await googleCloudCacheService.getEstimates(
@@ -61,7 +60,7 @@ export default class CacheManager implements EstimatorCache {
       return
     }
 
-    switch (this.currentCacheMode) {
+    switch (configLoader().CACHE_MODE) {
       case GCS:
         return googleCloudCacheService.setEstimates(estimates, grouping)
       default:
--- a/packages/app/src/EstimatorCacheGoogleCloudStorage.ts
+++ b/packages/app/src/EstimatorCacheGoogleCloudStorage.ts
@@ -6,7 +6,7 @@ import { Storage } from '@google-cloud/storage'
 import { Stream } from 'stream'
 import moment from 'moment'
 import EstimatorCache from './EstimatorCache'
-import { EstimationResult } from '@cloud-carbon-footprint/common'
+import { configLoader, EstimationResult } from '@cloud-carbon-footprint/common'
 import { EstimationRequest } from './CreateValidRequest'
 import { getCacheFileName } from './CacheFileNameProvider'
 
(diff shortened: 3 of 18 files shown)


packages/api/src/api.ts
```diff
@@ -112,12 +114,19 @@ const RecommendationsApiMiddleware = async function (
...
+  router.get('/regions/emissions-factors', EmissionsApiMiddleware)
+  router.get('/recommendations', RecommendationsApiMiddleware)
+  router.get('/healthz', (req: express.Request, res: express.Response) => {
+    res.status(200).send('OK')
+  })
+
+  return router
+}
+
+export default createRouter
```
dtoakley: Let's make this a named export on line 117, rather than the default export. This is to ensure naming consistency when importing and using the function. E.g. in the `server.ts` and test files.

packages/common/src/Logger.ts
```diff
@@ -23,24 +23,33 @@ enum LOGGING_LEVELS {
...
-      switch (configLoader().LOGGING_MODE) {
-        case 'GCP':
-          this.logger = this.getGCPLogger()
-          break
-        default:
-          this.logger = this.getDefaultLogger(logLabel)
+    this.logLabel = logLabel
+  }
+
+  get logger() {
```
dtoakley, line 37: I'm not sure we understand why this refactoring is required to allow you to use a different logger in the backstage plugin. It looks like it's just a refactor for now... Are these changes going to make using the backstage logger easier in some way?

packages/common/src/ConfigLoader.ts
```diff
@@ -2,8 +2,14 @@
  * © 2021 Thoughtworks, Inc.
  */
 
-import configFile, { CCFConfig } from './Config'
+import getConfig, { CCFConfig } from './Config'
+
+let _config: CCFConfig = undefined
+
+export const setConfig = (ccfConfig: CCFConfig = getConfig()) => {
```
dtoakley, line 9: So this new function makes sense - my only issue is at it now introduces another way that the config can be mocked in the tests. I think I'm okay with this small amount of tech debt for now, but we should mock consistently, probably using this new function across all our tests.

packages/app/src/EstimatorCacheGoogleCloudStorage.ts
```diff
@@ -33,6 +27,7 @@ export default class EstimatorCacheGoogleCloudStorage
     grouping: string,
   ): Promise<void> {
     const cachedEstimates = await this.getCloudFileContent(grouping)
+    const bucketName = configLoader().GCP.CACHE_BUCKET_NAME
```
dtoakley, line 30: I'm not quite following why refactoring the bucket name out of the constructor is needed for your use case. I'm also not sure it's an improvement... I think I find doing one call to get the config and pass it into the constructor easier to understand/maintain - but I might be missing something 🙃

packages/app/src/CacheManager.ts
```diff
@@ -34,7 +33,7 @@ export default class CacheManager implements EstimatorCache {
     const { GCS } = this.cacheModes
     let estimates
 
-    switch (this.currentCacheMode) {
+    switch (configLoader().CACHE_MODE) {
```
dtoakley, line 36: Similar thoughts as below in the `EstimatorCacheGoogleCloudStorage.ts` file

packages/gcp/src/domain/GcpFootprintEstimationConstants.ts
```diff
@@ -140,7 +140,7 @@ export const GCP_CLOUD_CONSTANTS: CloudConstantsByProvider = {
   SERVER_EXPECTED_LIFESPAN: 35040, // 4 years in hours
 }
 
-const getGCPEmissionsFactors = (): CloudConstantsEmissionsFactors => {
+export const getGCPEmissionsFactors = (): CloudConstantsEmissionsFactors => {
```
dtoakley, line 143: Let me know if I'm missing something, but this looks like a general refactor. I like it, but unless it's needed for the backstage plugin, I think a separate PR might be better. One reason is your branch still has a reference to `GCP_EMISSIONS_FACTORS_METRIC_TON_PER_KWH` in some test code, and I'd like to make sure the change is applied consistently.

packages/common/src/Config.ts
```diff
@@ -103,7 +103,7 @@ const getEnvVar = (envVar: string): string => {
   }
 }
 
-export const appConfig: CCFConfig = {
+const getConfig = (): CCFConfig => ({
```
dtoakley, line 106: FYI - this breaks out test coverage thresholds, but y'all don't need to worry about that. Our team can resolve any coverage issues.

packages/common/src/Logger.ts
HHogg: There were a few places around the code where the config was being used at the point the code was being executed. We've had to make these changes to delay the request for the config so that there is time during the setup of the router to set the config.

packages/app/src/EstimatorCacheGoogleCloudStorage.ts
HHogg: Again, same reply as above. The CacheManager was being instantiated at the time the code was being executed, which in turn was instantiated an EstimatorCacheGoogleCLoudStorage. Moving this out of the constructor means that it doesn't request the config until it's needed and allows the config to be set first.

packages/app/src/CacheManager.ts
HHogg: Same reason as above ☝️

packages/gcp/src/domain/GcpFootprintEstimationConstants.ts
HHogg: Same reason as above ☝️

HHogg: We'll update the other instances of `GCP_EMISSIONS_FACTORS_METRIC_TON_PER_KWH` (I thought we got them all!).

packages/common/src/Config.ts
HHogg: Sorry, we can write a test, that's not a problem.

packages/gcp/src/domain/GcpFootprintEstimationConstants.ts
tudi2d: Not sure here - do you refer to `GCP_EMISSIONS_FACTORS_METRIC_TON_PER_KWH` in `App.test.ts`?

packages/common/src/__tests__/Config.test.ts
```diff
@@ -0,0 +1,44 @@
+/*
+ * © 2022 Thoughtworks, Inc.
+ */
+
+import fs from 'fs'
+import getConfig from '../Config'
+
+describe('Config', () => {
```
dtoakley, line 8: Nice :)

packages/gcp/src/domain/GcpFootprintEstimationConstants.ts
dtoakley: Yeah, that is the place I'm referring to. Instead of mocking the removed constant, could we mock the return value for the new `getGCPEmissionsFactors` to be the same values: 

`gcpRegion1: 3,
    gcpRegion2: 4,`

packages/common/src/Logger.ts
dtoakley: Makes sense. Our team might look into refactoring that in the future, but that feel out of scope of the hack week

packages/common/src/__tests__/Config.test.ts
dtoakley: And thank you!

dtoakley:
Merged!

PR #734, opened 2022-03-08
Initial changes for backstage plugin

itura:
## Description of Change

Hacked together some changes which allow us to use CCF components in a backstage plugin.

Using this to have a convo about how to proceed with actual changes to the client

## Checklist

- [x] yarn test passes
- [x] yarn lint has been run

## Notes

© 2021 Thoughtworks, Inc.


packages/client/src/common/FilterBar/filterBarStyles.ts
```diff
@@ -10,8 +10,8 @@ const useStyles = makeStyles((theme) => ({
   filterHeader: {
     top: 0,
     left: 'auto',
-    position: 'fixed',
```
itura, line 13: this makes the filters render floating above the backstage header

packages/client/src/index.tsx
```diff
@@ -25,6 +25,5 @@ function Root() {
   )
 }
 
-ReactDOM.render(<Root />, document.getElementById('root'))
```
itura, line 28: this takes over the browser and overwrites everything backstage puts there. we will need to separate the component exports from rendering the app

packages/client/src/layout/ErrorPage/ErrorPage.tsx
```diff
@@ -21,13 +21,13 @@ type ErrorHandlingType = {
 }
 
 export const useErrorHandling = (): ErrorHandlingType => {
-  const navigate = useNavigate()
+  // const navigate = useNavigate()
```
itura: in the plugin we don't have access to the Router itself. 

we might be able to figure out a way around this, but backstage is oriented around propagating errors up into a centralized error boundary. while we could have some local error handling which does something like display a message / icon, it's not really in a backstage plugin's purview to navigate to arbitrary routes in local error handling

packages/client/src/pages/EmissionsMetricsPage/EmissionsMetricsPage.tsx
```diff
@@ -20,6 +20,9 @@ import useStyles from './emissionsMetricsStyles'
 import EmissionsSidePanel from './EmissionsSidePanel/EmissionsSidePanel'
 import { EmissionsFilters } from './EmissionsFilterBar/utils/EmissionsFilters'
 import { EstimationResult } from '@cloud-carbon-footprint/common'
+import {ThemeProvider} from "styled-components";
+
+const BASE_URL = "/api";
```
itura: in backstage plugins, you are supposed to rely on their built-in discovery mechanisms to find the API you're trying to talk to. see related changes in the remote service hooks.

there are different ways to slice this but the simplest is to inject the correct baseUrl. in our current hack, baseUrl is equal to `localhost:7007/api/cloud-carbon-footprint` which we retrieve from a [discovery mechanism](https://github.com/lcoffeejohnson/backstage/blob/cloud-carbon-footprint-plugin/plugins/cloud-carbon-footprint/src/components/ExampleComponent/ExampleComponent.tsx#L78)

itura:
@dtoakley @4upz-tw i think this is ready except for changeset, let me know your thoughts. happy to discuss the changes, especially the config stuff. see https://github.com/backstage/backstage/blob/91e82fbe17d4bd3d7f3c62b566bdef34be3a1a4a/plugins/cloud-carbon-footprint/src/components/CloudCarbonFootprintPlugin.tsx for what kind of usage these changes enable

PR #796, opened 2022-04-17
Integrate coefficients - CPU coefficients and Azure getMemory calculation

davidmytton:
## Description of Change

I implemented [the energy coefficients calculations](https://github.com/cloud-carbon-footprint/cloud-carbon-coefficients) in a separate repo because they make use of a Jupyter notebook. This made it easier to implement, but created a separation from the main project. For easier maintenance, they should probably be in the same repo. This PR is an initial attempt at starting that integration.

1. The first commit 2a209b1f646377e9a53b82191d2c8e7f52a7299a refactors where the CPU energy coefficients are located, removing the duplicate instances within each cloud provider package and moving them to the core package. All the CPU values are the same, so it seems to me they should be defined once. This will make it easier to keep them updated. I did they by creating a new type `ComputeProcessor` at https://github.com/cloud-carbon-footprint/cloud-carbon-footprint/commit/2a209b1f646377e9a53b82191d2c8e7f52a7299a#diff-ea0031aabb0814005caaf5aadd8e7f9b2d419e33bc3c5c01c0bbffe2c8123660R11
2. The second commit 628fb6cc8ec4c8f979d7939e5ab3d6819463c16e is a modified implementation of the `getMemory` function for Azure at https://github.com/cloud-carbon-footprint/cloud-carbon-footprint/commit/628fb6cc8ec4c8f979d7939e5ab3d6819463c16e#diff-b6bb7c806094085d26838e5a91bcbf8be822451342735057c4835fc4c4bcf472R43 which removes the hard coded `AZURE_CLOUD_CONSTANTS.MEMORY_AVG` and instead calculates it as a mean average of the listed CPUs, which is similar to how the hard coded value is calculated. This will make it easier to maintain as the list of supported CPUs changes. If this approach is acceptable then I will implement similar calculations across the packages with the aim that all values are calculated, not hard coded. This will eventually bring the SPECpower data into this repo to perform the CPU architecture calculations as well.

## Checklist

- [x] yarn test passes
- [x] yarn lint has been run

### MEMORY_AVG calculation

https://github.com/cloud-carbon-footprint/cloud-carbon-coefficients repo provides two different sets of averages:

- Memory average for a cloud provider: the mean average of all the SPECpower results where the CPU architecture matches the list of those being used by each cloud provider. This gives the individual cloud provider values e.g. for `AZURE_CLOUD_CONSTANTS.MEMORY_AVG`
- Memory average for CPU architecture: the mean average of all the SPECpower results with a specific CPU architecture. This gives the individual values for each CPU type e.g. `COMPUTE_PROCESSOR_TYPES.CASCADE_LAKE.gb_chip_avg` as implemented in this PR.

The implementation in 628fb6cc8ec4c8f979d7939e5ab3d6819463c16e removes the hard coded value (which is from the first average above) and replaces it with an average of the values in the second example. This is essentially an average of averages.

What we lose in accuracy, we gain in easier maintenance. You can see in the changed test values that the difference is quite small - https://github.com/cloud-carbon-footprint/cloud-carbon-footprint/commit/628fb6cc8ec4c8f979d7939e5ab3d6819463c16e#diff-3a91c4a5524debc1c2466b3fa5f8477aa20f77aeee3ff6a93f2ce23e1953dd94R994 - but there is still a difference.

### Draft

I'm opening this as a draft to request feedback on the approach before I do any more:

- Should the https://github.com/cloud-carbon-footprint/cloud-carbon-coefficients repo calculations be integrated? If you're happy with keeping them separate then I won't do any more!
- Does it make sense to move the CPU values to the core package?
- Does the `ComputeProcessor` type implementation make sense? I'm new to TypeScript so any feedback is welcome
- Any thoughts on the `MEMORY_AVG` calculation benefits/limitations above?


ccasher:
Hi @davidmytton, so sorry for such a delayed response here. Our team is behind both commits and planned implementation here. Please let us know how we can help to continue this work and progress it forward!

PR #853, opened 2022-06-29
Filter footprint on subscriptionIds (for Azure). Fixes #828

### LINKED ISSUE #828: Ability to filter response to API /footprint by subscription id (s)
When making a call /footprint in the API I would like to be able to send one or more Azure subscription ids as a parameter.
The response from /footprint should only contain results for the subscriptions specified in the input parameter
## Feature Suggestion
I work in a company that has a lot of Azure resources. I have an app registration that has access to 268 subscriptions.
My goal is to create a widget for Azure Devops that will display a graph of cloud carbon footprint.
The widget would call the Cloud Carbon Footprint API to get the data.
The problem with the current API is that it is really hard to start it with 268 subscriptions (there is a limit in Azure on how many requests you can make to get data) and after the timeout the request limit is hit again. The other problem is that it would take a long time to return the data.

**Describe the solution you'd like**
I would like a parameter to the /footprint-endpoint that contains the subscription ids that we want data from.
I could then specify the subscriptions when calling the API from the widget. After that the widget could display the cloud carbon footprint from the subscriptions that are relevant for a team. 
As a result a team that wants to know the cloud carbon footprint for their subscriptions could add the widget to a dashboard in their Azure Devops project (and specifying the subscription ids that are relevant for that team).

**Describe alternatives you've considered**
It would be possible to create an app registration that only has access to a limited number of subscriptions and then create an API instance that uses that app registration and then call that API instance from the widget.
That solution would not be optiomal for me since we have many teams and if all teams have to create Azure resources to be able to see the Cloud Carbon Footprint for their subscriptions that would take a lot of effort but also increase the Cloud Carbon Footprint that we would like to reduce.

**Additional context**

## Possible Implementation

- Add subscriptionIds to FootprintEstimatesRawRequest (could be comma separated)
- Add subscriptionIds to EstimationRequest
- Modify App.ts so that subscriptionIds are sent to AzureAccount.getDataFromConsumptionManagement
- Modify getDataFromConsumptionManagement so that we only call getDataForSubscription if the subscription is in subscriptionIds

I realize that this suggestion will only work for Azure resources.

nilshedstrom:
## Description of Change
Added ability to specify the footprint requests to specific subscriptionIds (comma separated string) that is sent as a part of the request to the API.

I also needed update the caching so that it does not use the same cache file for different subscriptionIds.

Fixes #828 

## Checklist

- [X] tests are changed or added
- [ ] yarn test passes 
- [ ] yarn lint has been run (it fails for files that I did not change)

## Notes

© 2021 Thoughtworks, Inc.

### DIFF
--- a/packages/api/src/api.ts
+++ b/packages/api/src/api.ts
@@ -41,6 +41,7 @@ const FootprintApiMiddleware = async function (
     endDate: req.query.end?.toString(),
     ignoreCache: req.query.ignoreCache?.toString(),
     groupBy: req.query.groupBy?.toString(),
+    subscriptionIds: req.query.subscriptionIds?.toString()
   }
   apiLogger.info(
     `Footprint API request started with Start Date: ${rawRequest.startDate} and End Date: ${rawRequest.endDate}`,
--- a/packages/app/src/App.ts
+++ b/packages/app/src/App.ts
@@ -42,6 +42,7 @@ export default class App {
     const endDate = request.endDate
     const grouping =
       (request.groupBy as GroupBy) || configLoader().GROUP_QUERY_RESULTS_BY
+    const subscriptionIds = request.subscriptionIds;
     const config = configLoader()
     const AWS = config.AWS
     const GCP = config.GCP
@@ -117,6 +118,7 @@ export default class App {
           startDate,
           endDate,
           grouping,
+          subscriptionIds
         )
         appLogger.info('Finished Azure Estimations')
         AzureEstimatesByRegion.push(estimates)
--- a/packages/app/src/Cache.ts
+++ b/packages/app/src/Cache.ts
@@ -83,7 +83,7 @@ export default function cache(): any {
         )
         cacheLogger.info('Setting new estimates to cache file...')
         if (estimatesToPersist.length > 0) {
-          await cacheManager.setEstimates(estimatesToPersist, grouping)
+          await cacheManager.setEstimates(estimatesToPersist, grouping, request.subscriptionIds)
         }
       }
 
--- a/packages/app/src/Cache.ts
+++ b/packages/app/src/Cache.ts
@@ -171,6 +171,7 @@ function getMissingDataRequests(
       ignoreCache: false,
       groupBy: request.groupBy,
       region: request.region,
+      subscriptionIds: request.subscriptionIds
     }
   })
 }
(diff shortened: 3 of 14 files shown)


4upz (commented):
Hi @nilshedstrom, my apologies for the delay in getting back to you and thanks for working on this! I just got done reviewing your PR and had a few notes/suggestions:

1. I would recommend not introducing dynamic naming/hashing for the caching methods. In addition, the CacheManagers should not have the subscription IDs passed to them as we'd like them to not be aware of any filtering methods. This is because we're currently working on major cache improvements that will conflict with this, and because it is also better to have the cache save all estimates consistently and only filtering them when it's time to retrieve them. I would recommend leaving out the changes until we're done with our cache improvements, or having a quick fix to only have the getEstimates function be aware of the IDs so that it can filter the results by accountId when retrieving estimates from the cache.
2. After you make the changes to the cache logic, we can build upon your PR to make providing account IDs extensible to other cloud providers. Feel free to let us know when you're done with the cache changes.
3.  If you'd like, you're welcome to contribute the extensibility refactor that we're thinking. This would involve making a new POST method for Footprint API endpoint, and passing the parameters as a body instead. This would allow for additional filters to eventually be added (another item in our backlog) and for the the passing of IDs to appear as a JSON body in this format:
``` json
{ 
	"start": "2021-09-12",
	"end": "2021-11-14",
	"ignoreCache": true,
	"groupBy": "month",
	"accountIds":  {
		"AWS": ["1", "2", "3"],
		"GCP": ["1", "2", "3"],
		"AZURE": ["1", "2", "3"]
	}
}
``` 
Let me know if you're interested in this, and I can provide additional context.

Hopefully that's clear. Feel free to reply with any questions or concerns, and thanks again! 😁

4upz:
Hi @nilshedstrom just wanted to follow up on this. Thanks again for tackling this solution! We've recently implemented some significant cache changes, including our new option to use a MongoDB instance as a cache. These changes brought along a beginning implementation for API filtering by account IDs (including Azure subscriptions) for all three cloud providers. You can learn more [here](https://www.cloudcarbonfootprint.org/docs/data-persistence-and-caching#filtering-estimates). Since these changes conflict with that approach and this PR has gotten stale, I'm going to go ahead and close this.

However, feel free to pull the most recent changes and tackle a solution for making these filter params work for the other cache methods! Hopefully our newest changes and the potential for more filtering proves to be helpful for your use case.

PR #904, opened 2022-09-01
Aggregate footprint estimates for AWS resources by tags 

charleskorn:
## Description of Change

Adds the ability to aggregate estimates based on resources' tags.

## Checklist

- [x] yarn test passes
- [x] yarn lint has been run
- [ ] (n/a) relevant documentation is changed or added

## Notes

Currently only supports AWS, and only when Cost and Usage Reporting is the data source.

I haven't added any filtering to the UI - this is an API-only change.

This is a draft PR - open to feedback and suggestions.

© 2021 Thoughtworks, Inc.


ghost (approved):
Changes on TZ support look ok to me.

packages/app/src/__tests__/App.test.ts
```diff
@@ -187,6 +187,7 @@ describe('App', () => {
                 cost: 5,
                 region: region,
                 usesAverageCPUConstant: false,
+                tags: {},
```
ccasher: I believe we could make tags as an optional key here in the EstimationResult type. that way we wouldn't have to update all these test.

packages/aws/src/__tests__/AWSAccount.test.ts
```diff
@@ -596,6 +596,7 @@ function getExpectedEstimationResult(startDate: Date = new Date()) {
           cost: 3,
           region: region,
           usesAverageCPUConstant: false,
+          tags: {},
```
ccasher: Same as above

packages/aws/src/__tests__/CostAndUsageReports.test.ts
```diff
@@ -176,6 +180,10 @@ describe('CostAndUsageReports Service', () => {
             serviceName: 'AmazonEC2',
             cost: 3,
             region: 'us-east-1',
+            tags: {
```
ccasher, line 191: Same as above

packages/aws/src/lib/CostAndUsageReports.ts
```diff
@@ -93,21 +93,24 @@ export default class CostAndUsageReports {
...
     start: Date,
     end: Date,
     grouping: GroupBy,
   ): Promise<EstimationResult[]> {
-    const usageRows = await this.getUsage(start, end, grouping)
-    const usageRowsHeader: Row = usageRows.shift()
+    const tagNames = configLoader().AWS.RESOURCE_TAG_NAMES
+    const usageRows = await this.getUsage(start, end, grouping, tagNames)
+
+    usageRows.shift()
```
ccasher, line 107: Could you please explain the rationale for this change?

packages/aws/src/lib/CostAndUsageReports.ts
```diff
@@ -145,34 +150,18 @@ export default class CostAndUsageReports {
     const unknownRows: CostAndUsageReportsRow[] = []
 
     inputData.map((inputDataRow: LookupTableInput) => {
```
ccasher, line 162: Will tagging support affect this LUT code?

packages/aws/src/lib/CostAndUsageReports.ts
```diff
@@ -527,30 +516,37 @@ export default class CostAndUsageReports {
     start: Date,
     end: Date,
     grouping: GroupBy,
+    tagNames: string[],
```
ccasher, line 540: If possible, could you please add in comments to explain the rationale behind these changes? i am assuming there is a specific way that Athena is dealing with tags

packages/aws/src/lib/CostAndUsageReportsRow.ts
```diff
@@ -22,34 +22,40 @@ import { concat } from 'ramda'
...
-    ).flat()
-    const billingDataRowValues = rowData
-      .map((column) => Object.values(column))
-      .flat()
-    const billingDataRow = Object.fromEntries(
-      billingDataRowKeys.map((_, i) => [
-        billingDataRowKeys[i],
-        billingDataRowValues[i],
-      ]),
+  constructor(rowData: Athena.datumList, tagNames: string[]) {
```
ccasher: Please provide rationale for this change

packages/azure/src/__tests__/ConsumptionManagement.test.ts
```diff
@@ -128,6 +128,7 @@ describe('Azure Consumption Management Service', () => {
             serviceName: 'Virtual Machines',
             cost: 5,
             region: 'uksouth',
+            tags: {},
```
ccasher: Could be reverted if tags are optional

packages/azure/src/lib/ConsumptionDetailRow.ts
```diff
@@ -26,6 +26,7 @@ export default class ConsumptionDetailRow extends BillingDataRow {
     this.vCpuHours = this.usageAmount * this.getVCpus()
     this.gpuHours = this.usageAmount * this.getGpus()
     this.region = this.getRegionFromResourceLocation()
+    this.tags = {} // TODO: we currently don't support resource tags for Azure
```
ccasher: Same as above

packages/azure/src/lib/ConsumptionManagement.ts
```diff
@@ -132,6 +132,7 @@ export default class ConsumptionManagementService {
             consumptionDetailRow,
             footprintEstimate,
             grouping,
+            [],
```
ccasher, line 135: Same as above

packages/common/src/EstimationResult.ts
```diff
@@ -25,6 +25,11 @@ export interface ServiceData {
   accountName: string
   serviceName: string
   region: string
+  tags: TagCollection
```
ccasher: here is where you could update to:
`tags?: TagCollection`

packages/common/src/helpers/__tests__/helpers.test.ts
```diff
@@ -67,7 +64,10 @@ describe('Helpers', () => {
       [GroupBy.quarter, new Date('2020-07-01T00:59:59.000Z')],
       [GroupBy.year, new Date('2021-03-31T23:59:59.000Z')],
     ]).it('should get period end date for %s', (grouping, expectedResult) => {
-      const result = getPeriodEndDate(new Date(), grouping)
+      const result = getPeriodEndDate(
```
ccasher: Some of these changes look like the ones submitted in a separate PR. Would it be possible to remove these and keep them separated?

packages/common/src/helpers/helpers.ts
```diff
@@ -28,19 +28,18 @@ export const getHoursInMonth = (): number => {
...
 
-//01/01- 00 --01/01 -23-59-59
-
 export const getPeriodEndDate = (startDate: Date, grouping: GroupBy): Date => {
   const periodGrouping: { [key: string]: Date } = {
-    day: moment(startDate).add(1, 'd').subtract(1, 's').toDate(),
-    week: moment(startDate).add(1, 'w').subtract(1, 's').toDate(),
-    month: moment(startDate).add(1, 'M').subtract(1, 's').toDate(),
-    quarter: moment(startDate)
+    day: moment.utc(startDate).add(1, 'd').subtract(1, 's').toDate(),
```
ccasher: same as above

packages/core/src/FootprintEstimate.ts
```diff
@@ -82,6 +86,7 @@ export interface MutableServiceEstimate {
   cost: number
   region: string
   usesAverageCPUConstant: boolean
+  tags: TagCollection
```
ccasher: make optional for now

packages/core/src/__tests__/CloudProviderAccount.test.ts
```diff
@@ -96,6 +96,7 @@ describe('CloudProviderAccount', () => {
             kilowattHours: 1000,
             serviceName: 'test-service',
             usesAverageCPUConstant: true,
+            tags: {},
```
ccasher: can revert ones tags are optional

packages/common/src/helpers/__tests__/helpers.test.ts
charleskorn: Yep - once #903 is merged I'll rebase. (I needed this fix on this branch to make sure these changes hadn't broken any other tests.)

packages/app/src/__tests__/App.test.ts
charleskorn: Sure, makes sense.

packages/aws/src/lib/CostAndUsageReports.ts
charleskorn: Both use `CostAndUsageReportsRow`, and `CostAndUsageReportsRow` assumes its data is coming from Athena. So we need to rejig what's happening here to match the changes to `CostAndUsageReportsRow` that were needed to support tagging (pending discussion on these changes below).

packages/aws/src/lib/CostAndUsageReports.ts
charleskorn: Let's discuss in https://github.com/cloud-carbon-footprint/cloud-carbon-footprint/pull/904/files/19dd86f43a82d6ea7f4f8280323ad4f42f02cf5f#r971225231

packages/aws/src/lib/CostAndUsageReports.ts
charleskorn: Sure. Let me know if this is clear and then I'll add it in:

> AWS' Cost and Usage Reporting (CUR) translates tags' names to names that are valid Athena column names. On top of this, it also adds a prefix to distinguish between user-created tags and AWS-internal tags.
> This isn't documented anywhere, but the behaviour appears to be that a tag such as `SourceRepository` will be `user:SourceRepository` in CUR, and `resource_tags_user_source_repository` in Athena. (AWS-internal tags will be prefixed with `aws:` instead of `user:` in CUR.)

We then need to add these column names to the query (so that Athena returns their values) and to the `GROUP BY` clause (so that the values are aggregated by the tag values).

Looking at this again with fresh eyes, I'm not sure why I continued with using indices (like the existing code) rather than names for the column names in the `GROUP BY` - I'll try swapping over to column names to make it clearer.

The rest of the changes are formatting changes or variable extraction to make the query easier to read (all the extra linebreaks created by Prettier made it difficult to follow, at least for me).

packages/aws/src/lib/CostAndUsageReportsRow.ts
charleskorn: The main thing that motivated this change is that Athena returns each row (`rowData` here) as an array like this:

```json
[
  { "VarCharValue": "2022-09-15T00:00:00.0000Z" },
  { "VarCharValue": "my-account" },
  { "VarCharValue": "EC2" },
  { "VarCharValue": "us-east-1" },
  { "VarCharValue": "xxx" },
  { "VarCharValue": "seconds" },
  { "VarCharValue": "2.0" },
  { "VarCharValue": "Production" },
  { "VarCharValue": "3.0" },
  { "VarCharValue": "5.0" }
]
```

If the constructor was to use the original mechanism (turning the row into an object with properties matching the column names in the Athena result), then handling tags would be a little awkward: the tags would get applied to the object with property names like `resource_tags_user_environment`, and then we'd have to do some extra work to pull them out of either the row or the object to put them in `tags`.
  
Directly taking the tag values from the row seemed easier and clearer rather than continuing with the original mechanism.
  
Very open to feedback or other suggestions though! Another option could be to pass in the property values to the constructor. Then all of the knowledge of Athena and the query structure could live in `CostAndUsageReports`, and this class becomes much simpler and has no knowledge of Athena. This would also make the LUT handling easier: we wouldn't need to convert the LUT data into an Athena-like format just for `CostAndUsageReportsRow` to then transform it again.

packages/aws/src/lib/CostAndUsageReports.ts
charleskorn: `GROUP BY` now uses column names rather than indices: https://github.com/cloud-carbon-footprint/cloud-carbon-footprint/pull/904/commits/026636b7cf203305ee6ce333a3d4828f6614a50c

packages/aws/src/lib/CostAndUsageReports.ts
```diff
@@ -677,3 +674,37 @@ export default class CostAndUsageReports {
...
+      }
+
+      columnName = columnName + char.toLowerCase()
+    }
+  }
+
+  return columnName
+}
+
+const isUpperCase = (text: string): boolean => {
```
ccasher, line 770: These functions could potentially live in a common / helpers directory rather than the CUR file

charleskorn: True - if there's a need for them elsewhere, we could move them then.

packages/aws/src/lib/CostAndUsageReports.ts
charleskorn: Comment has been added in [`1c66593`](https://github.com/cloud-carbon-footprint/cloud-carbon-footprint/pull/904/commits/1c6659328c923487338cd66708696ddb41837eab).

packages/aws/src/lib/CostAndUsageReportsRow.ts
camcash17: hey Charles, thanks for the explanation. Would it be possible to refactor to pass through the properties to the constructor. I think this direction might be a bit cleaner. The logic for athena could even be extracted out into its own file that is imported to the CUR file. And some of the functions mentioned above could also live there.

packages/core/src/BillingDataRow.ts
```diff
@@ -13,13 +15,14 @@ export default abstract class BillingDataRow {
...
   public timestamp: Date
-  public vCpus: number
+  public vCpus?: number
   public vCpuHours: number
   public gpuHours: number
   public machineType: string
   public seriesName: string
   public instanceType: string
   public replicationFactor: number
+  public tags: TagCollection
```
camcash17: i think this needs to be marked as optional

packages/aws/src/lib/CostAndUsageReportsRow.ts
charleskorn: Resolved in https://github.com/cloud-carbon-footprint/cloud-carbon-footprint/pull/904/commits/f98ca8cafc47a51148ab5974a44bc84cd3ec2058

packages/core/src/BillingDataRow.ts
charleskorn: Resolved in https://github.com/cloud-carbon-footprint/cloud-carbon-footprint/pull/904/commits/4b2923d0de73dff9e9473a56b819709069a02ec3.

camcash17 (approved):

PR #926, opened 2022-10-04
Update AWS S3 Standard replication factor

### LINKED ISSUE #755: Update AWS S3 Standard replication factor
Update the AWS S3 Standard replication factor

## Feature Update
Today both S3 Standard and S3 Standard IA use the same replication factor (3) however S3 standard offers a 99.99% availability when Standard IA offers a 99.9% availability.

### Context
Our hypothesis is that for S3 Standard data is replicated in 2 other AZs but it is also mirrored locally (in each AZ) to achieve better availability in case of failure (avoid having to reach another AZ).

We could consider that in that case the replication factor is 6 (3 AZs + mirror).

**Tech Tasks**
- [ ] Update `packages/aws/src/domain/AwsFootprintEstimationConstants.ts` to include a replication factor of *6* for  standard S3 services.
- [ ] Run and update the necessary tests to ensure that they reflect the new replication factor

akhil-ghatiki:
## Description of Change

Updated AWS S3 Standard replication factor

Updated the replication factor to 6 and modified the test cases accordingly to match the new replication factor.
Attempt to fix : #755 
@github-benjamin-davy @4upz 

## Checklist

- [x] yarn test passes
- [x] yarn lint has been run

© 2021 Thoughtworks, Inc.


4upz (approved):
Awesome, thanks for again for contributing! 🎃

PR #988, opened 2022-11-23
Integration with carbon aware

### LINKED ISSUE #973: Time forecast and cloud region recommendation
## Feature Suggestion
**The problem:**
The feature is an addition to the existing tool. Data centres and servers are expected to create 14 % of the world's emissions by 2040. Thus, it's significant for us to use energy when it's being produced by renewable resources and/or from locations that have lesser carbon emission.

**The solution:**
There are two functionality in this feature:
1. Time forecast: provides optimal time to schedule workloads in next 24 hours of timeframe
2. Cloud region recommendation: retrieves all neighbouring cloud region for a particular account and recommends a region that has the least carbon emission based on data of last 1 month

Note: We'll be implementing the above mentioned functionality as a feature flag that can be enabled and disabled through environment variables.

Note: It makes use of [carbon aware sdk](https://github.com/Green-Software-Foundation/carbon-aware-sdk) endpoints to fetch data related to the cloud regions.

**Screenshots:**
1. Time forecast
Screenshot 2022-11-04 at 11 01 42 AM
Screenshot 2022-11-04 at 11 02 25 AM

2. Region recommendation
Screenshot 2022-11-04 at 11 03 18 AM

tanvi-sutar:
Fixes #973 issue


jest.base.config.js
```diff
@@ -10,6 +10,7 @@ module.exports = {
   coverageReporters: ['json', 'lcov', 'text', 'clover'],
   collectCoverageFrom: ['<rootDir>/src/**/*.ts'],
   setupFiles: ['../../setupTests.ts'],
+  preset: 'ts-jest',
```
4upz, line 13: Is there a reason behind this config change?

package.json
```diff
@@ -51,6 +52,7 @@
     "docs": "docsify serve --port 3001",
     "release": "changeset version && yarn install",
     "postinstall": "husky install",
-    "version": "yarn changeset version && yarn install --no-immutable"
+    "version": "yarn changeset version && yarn install --no-immutable",
+    "prepare": "husky install"
```
4upz, line 59: Could you explain these changes, along with the babel dependency?

packages/client/.env.template
```diff
@@ -22,3 +22,8 @@ REACT_APP_DISABLE_CACHE=true
 # MongoDB config
 # When using MongoDB as cache, large results will be paginated and split into multiple requests. Use this to set a limit for results per page/request.
 PAGE_LIMIT=30
+
+#variables for enabling and disabling region recommendation and forecast feature
+REACT_APP_ENABLE_REGION_RECOMMENDATION_FEATURE = true
+REACT_APP_ENABLE_FORECAST_FEATURE = true
+REACT_APP_CARBON_AWARE_SDK_BASE_URL = http://localhost:5073
```
4upz, line 29: I believe these variables make sense, but should probably live in the API/backend along with the SDK calls. 

For `REACT_APP_ENABLE_FORECAST_FEATURE`, this should be renamed or made more specific so that it's not confused with the Forecast component on the recommendations page. 

In regards to, `REACT_APP_CARBON_AWARE_SDK_BASE_URL` does this require a locally or self-hosted version of the SDK? If so, we should probably have documentation on how to set this up for CCF

packages/client/__mocks__/styleMock.js
```diff
@@ -0,0 +1 @@
+module.exports = {}
```
4upz, line 1: Is this necessary for client tests?

packages/client/babel.config.js
```diff
@@ -0,0 +1,7 @@
+module.exports = {
+  presets: [
+    '@babel/preset-react',
+    ['@babel/preset-env', { targets: { node: 'current' } }],
+    '@babel/preset-typescript',
+  ],
```
4upz, line 6: Not sure if these are necessary

packages/client/__mocks__/fileMock.js
```diff
@@ -0,0 +1 @@
+module.exports = 'test-file-stub'
```
4upz, line 1: Unnecessary -- if a test needs this, it should be a mock local to that test file

packages/client/jest.config.js
```diff
@@ -0,0 +1,9 @@
+module.exports = {
+  preset: 'ts-jest',
+  testEnvironment: 'jsdom',
+  moduleNameMapper: {
+    '\\.(jpg|jpeg|png|gif|eot|otf|webp|svg|ttf|woff|woff2|mp4|webm|wav|mp3|m4a|aac|oga)$':
+      '<rootDir>/__mocks__/fileMock.js',
+    '\\.(css|sass)$': '<rootDir>/__mocks__/styleMock.js',
+  },
+}
```
4upz, line 9: I don't think this is necessary but open to discussing

packages/client/package.json
```diff
@@ -104,6 +107,7 @@
     "@types/styled-components": "^5.1.9",
     "@typescript-eslint/eslint-plugin": "^5.9.0",
     "@typescript-eslint/parser": "^5.9.0",
+    "babel-jest": "^29.2.2",
```
4upz, line 107: May not need these babel dependencies

packages/client/src/common/AccountFilter/AccountFilter.tsx
```diff
@@ -1,8 +1,8 @@
-/*
- * © 2021 Thoughtworks, Inc.
- */
+// /*
+//  * © 2021 Thoughtworks, Inc.
+//  */
```
4upz, line 3: Please remove extra comment markers (`//`)

packages/client/src/pages/CarbonHackCards/ForecastCard/ForecastCard.test.tsx
```diff
@@ -0,0 +1,80 @@
+import React from 'react'
+import { render, within } from '@testing-library/react'
+import ForecastCard from './ForecastCard'
+import { mockData } from '../../../utils/data/mockData'
+import useRemoteForecastService from '../../../utils/hooks/ForecastServiceHook'
+import { forecastServiceMockReturnValue } from './ForecastCardMockData'
+import { ServiceResult } from '../../../Types'
+import { ForecastResult } from '../../../utils/hooks/ForecastServiceHook'
+import { BrowserRouter } from 'react-router-dom'
```
4upz, line 9: Let's add a new line to separate the imports from the begging in the variable declarations

packages/client/src/pages/EmissionsMetricsPage/CarbonIntensityMap/CarbonIntensityMap.test.tsx
```diff
@@ -5,6 +5,7 @@
 import CarbonIntensityMap from './CarbonIntensityMap'
 import { fireEvent, render, within } from '@testing-library/react'
 import { create } from 'react-test-renderer'
+import React from 'react'
```
4upz, line 8: I don't think this is needed

packages/client/src/pages/CarbonHackCards/RegionRecommendationCard/CloudRegions.ts
```diff
@@ -0,0 +1,176 @@
...
+  US_WEST_2_STAGE: { name: 'westus2stage', options: ['westus2stage'] },
+  US_WEST_3: { name: 'westus3', options: ['westus3', 'uswest3'] },
+  BRAZIL: { name: 'brazil', options: ['brazil'] },
+  BRAZIL_SOUTH: { name: 'brazilsouth', options: ['brazilsouth'] },
+  BRAZIL_SOUTH_EAST: { name: 'brazilsoutheast', options: ['brazilsoutheast'] },
+
+  ALL: { name: 'All Regions', options: ['All Regions'] },
+  UNASSIGNED: { name: 'Unassigned', options: ['Unassigned'] },
+  UNKNOWN: { name: 'Unknown', options: ['Unknown'] },
+}
```
4upz, line 176: These regions are already constants within each respective Cloud Provider package and shouldn't be copied over. I think this is another good reason why the SDK logic should live in the API and App packages if it depends on that information from the Cloud Providers

packages/client/src/pages/CarbonHackCards/ForecastCard/ForecastLineChartMockData.ts
```diff
@@ -0,0 +1,14 @@
...
+
+export const categories = [
+  '2022-11-08 15:40:00',
+  '2022-11-08 15:45:00',
+  '2022-11-08 15:50:00',
+  '2022-11-08 15:55:00',
+  '2022-11-08 16:00:00',
+  '2022-11-08 16:05:00',
+  '2022-11-08 16:10:00',
+]
```
4upz, line 14: I think this is small enough to live in the test file that requires it

packages/client/src/pages/CarbonHackCards/ForecastCard/ForecastLineChart.tsx
```diff
@@ -0,0 +1,96 @@
+import React from 'react'
+import ReactApexChart from 'react-apexcharts'
+
+export class ForecastLineChart extends React.Component<
+  // eslint-disable-next-line @typescript-eslint/ban-types
+  { data: object; categories: object },
+  { series: object[]; options: object }
```
4upz, line 7: These types can be a little more specific or just be `any` if they're expected props of apex charts

packages/client/src/pages/CarbonHackCards/ForecastCard/ForecastLineChart.tsx
```diff
@@ -0,0 +1,96 @@
+import React from 'react'
+import ReactApexChart from 'react-apexcharts'
+
+export class ForecastLineChart extends React.Component<
```
4upz, line 4: This should be a functional component instead of a class component. There should be a different way to construct the chart config without the use of constructors, similar to how we do for the other components

packages/client/src/utils/hooks/ForecastDataHook.ts
```diff
@@ -0,0 +1,48 @@
...
+  }
+
+  const { data, error, loading } = useRemoteForecastService(promiseArray)
+  let index = 0
+  const result = new Map()
+  for (const [key] of params) {
+    result.set(key, data[index++])
+  }
+  return { result, error, loading }
+}
```
4upz, line 48: We should use our API as a "middle-ma" to making calls to the Carbon Aware SDK. The client should not have an additional service to rely on besides the CCF API. I'd recommend looking into refactoring this in a way to where SDK calls, configs, and any non-presentational logic occur through the API, and this Data/Service hook (which can be combined into one, not sure if separate ones are needed) only handles making the call to the CCF endpoint.

Since it seems that the forecast logic also relies on emissions data, perhaps the fetching and formatting of the data can occur whenever estimates are initially fetch, and have it be included in the results? Since it relies on service estimates, this may cause problems on the client when dealing with large scale data. Open to discussing

packages/client/src/pages/EmissionsMetricsPage/EmissionsMetricsPage.tsx
```diff
@@ -80,6 +80,15 @@ export default function EmissionsMetricsPage({
...
               />
+              {process.env.REACT_APP_ENABLE_REGION_RECOMMENDATION_FEATURE ===
+              'true' ? (
+                <RegionRecommendationCard
+                  data={footprint.filteredData}
+                ></RegionRecommendationCard>
+              ) : null}
+              {process.env.REACT_APP_ENABLE_FORECAST_FEATURE === 'true' ? (
+                <ForecastCard data={footprint.filteredData}></ForecastCard>
+              ) : null}
```
4upz, line 91: `env` variables should be referenced using the `loadConfig` function (and any necessary default values should also be set there).

4upz (commented):
Thanks for opening this PR @tanvi-sutar! Apologies on the delay in getting back with feedback, as there is a lot here to review! I left a few comments ahead of our follow-up call. While this feature is awesome, I'm not sure if it's quite PR ready. To sum up my concerns: 

- Folder structure is still under Carbon Hack. This should be named differently, and also as a subfolder underneath the EmissionsMetricsPage folder since that's where it looks like these new components are rendered.
- All of the logic for this feature exists on the client, which may not be a good idea for performance reasons. We should discuss how this can live in the backend through the `App` and `API` packages, and how to offload the logic and calls from the client
- There seems to be quite a bit of extra dependencies that are added and changes to the jest configs that I'm not sure are needed
- Since this is such a huge feature, and it looks like it requires some setup for the Carbon Aware SDK to be self-hosted, there should be some documentation included on what this feature fully does and how to set it up (along with any specific instructions on the Carbon Aware SDK).

Issue #877, opened 2022-07-29
Export and Relocate Compute Optimizer CloudFormation only works with us-west-2


choweiyuan:
### Preflight Checklist

- [x] I have read the [Contributing Guidelines](https://github.com/cloud-carbon-footprint/cloud-carbon-footprint/blob/trunk/CONTRIBUTING.md) for this project.
- [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details
AWS - CloudFormation, Lambda, IAM

ccf-compute-optimizer cloudformation seems to have a condition to only deploy in USWest2 (Ohio). 
https://github.com/cloud-carbon-footprint/cloud-carbon-footprint/blob/trunk/cloudformation/ccf-compute-optimizer.yaml#L23-L26
https://github.com/cloud-carbon-footprint/cloud-carbon-footprint/blob/trunk/cloudformation/ccf-compute-optimizer.yaml#L66-L122

Is there a particular reason why Ohio must be the region to deploy the central bucket and the associated lambda functions? Can I deploy it elsewhere instead, e.g. Europe? Sure enough I could modify the template myself and then deploy without issues, but would be good to understand the reasoning behind this, and make this more flexible (not region restricted)

### Expected Behavior
I would like to deploy central bucket + compute optimizer recommendation lambda on a different region eg European regions
via CloudFormation (ccf-computer-optimizer.yaml)
### Actual Behavior
CloudFormation wouldn't work with other regions as the condition isnt met (only works with us-west-2)

### To Reproduce
Run ccf-compute-optimizer.yaml anywhere other than Ohio.

### Additional Information

© 2021 Thoughtworks, Inc.

jeff-zinger:
Hi there @choweiyuan! Thanks for reporting this issue. We're going to pick it up with the intention of allowing the deployment to different regions.

PR #989, opened 2022-11-25
Enable compute optimizer infra deployments from all regions

jeff-zinger:
In response to #877 , we are removing the requirement of only deploying the infrastructure for the compute optimizer from the us-west-2 region.


dfratte (approved):
I've tested this against regions in the US and EU. It deploys fine. Waiting for maintainer's approval before merge.

ccasher (approved):

PR #999, opened 2022-12-12
[doc] Fix wrong indent in docker usage snippet

Joffref:
Indeed, as visible on the website, docker snippet is not indent correctly. It causes trouble when copy and paste.

I've reindent the snippet. 

## Checklist


4upz (approved):
Thanks for catching this @Joffref! I'll go ahead and merge in your fix 🎉

PR #1008, opened 2023-01-08
Add B DTU storage type to Azure list

barryokane:
## Description of Change

Ref: https://www.cloudcarbonfootprint.org/docs/azure#unsupported-usage-types

Received the following message:

> [ConsumptionManagement] warn: Unexpected usage type for storage service: B DTU

This PR fixes that by adding "B DTU" to packages\azure\src\lib\ConsumptionTypes.ts

## Notes

Quick search seems to imply that both "B DTU" and "B DTUs" are possible, so I have left them both in.

© 2021 Thoughtworks, Inc.


4upz:
Awesome, looks good, @barryokane! Thanks for taking the time to add this! I'll go ahead and merge it in

PR #1020, opened 2023-01-25
[964] Add azure tags and resource group to estimates

### LINKED ISSUE #964: Add support for Azure Tags
#### Overview
We currently have the ability to make tags visible for CCF API results as detailed in our [Tagging Documentation](https://www.cloudcarbonfootprint.org/docs/tagging). Currently, this solution is only supported in AWS and we'd like to bring this functionality for Azure estimates as well. The goal is to modify the consumption management API query for Azure and other relevant areas in the `Azure` package to include both `resourceGroups` and `tags` for every resource in the `tags` field of the resulting `EstimationResult`. 

To keep consistent with the AWS implementation, desired tags should be specified via environment variable in the `.env` file of the `API` or `CLI` package. The variable should be named `AWS_RESOURCE_TAG_NAMES` and take an array of strings representing each tag name to include.

_Note:_ It may be worth considering adding a value for this config variable that can be used to include all tags for a resource.

#### Out of Scope:

- Adding any filtering for Azure tags outside of existing filters
- Displaying tags on the UI
- Adding tags for other packages (i.e. GCP)

#### Related issues/Dependencies:
- AWS Implementation: https://github.com/cloud-carbon-footprint/cloud-carbon-footprint/pull/904
- Original API Filtering (includes existing tag filters): https://github.com/cloud-carbon-footprint/cloud-carbon-footprint/issues/911

## Tech Tasks
- [ ] Add appropriate test coverage, including edge cases.
- [ ] Did you make changes to the client, api or cli package? If so, considering updating the create-app templates. You can use the `./scripts/update_templates.sh [package-name]` command from the root of the package to do so.
- [ ] If any create-app templates have been updated, make sure to run the [test-create-app script](https://docs.google.com/document/d/1Ei6Cb2DjwEJUNtpULJSxbOuIdAfCARERgE3BSsKWrsc/edit) to ensure it still works properly.

## Review Tasks
- [ ] Add Azure-specific tagging documentation to microsite (for CCF team)

NicolasBissig:
## Description of Change

fixes #964 

This PR adds support for azure tags and resourceGroups via the existing `tags` field in the estimations as described in #964 

## Checklist

- [X] tests are changed or added
- [ ] yarn test passes
      (There are problems with some tests from the azure recommendations, but I did not change any values there...)
- [X] yarn lint has been run
- [X] yarn changeset was run and relevant packages have been updated as a major/minor/patch bump with descriptions
      (Was out of scope)

## Notes

No points in the `Out of scope` section of the issue have been resolved yet.

© 2021 Thoughtworks, Inc.

### DIFF
--- a/.changeset/quiet-dots-relax.md
+++ b/.changeset/quiet-dots-relax.md
@@ -1,0 +1,5 @@
+---
+'@cloud-carbon-footprint/azure': minor
+---
+
+Add support for Azure Tags and resource groups
--- a/.changeset/quiet-dots-relax.md
+++ b/.changeset/quiet-dots-relax.md
@@ -1,5 +1,7 @@
 ---
 '@cloud-carbon-footprint/azure': minor
+'@cloud-carbon-footprint/api': minor
+'@cloud-carbon-footprint/common': minor
 ---
 
 Add support for Azure Tags and resource groups
--- a/packages/api/.env.template
+++ b/packages/api/.env.template
@@ -60,6 +60,9 @@ AZURE_TENANT_ID=your-azure-tenant-id
 # Optionally set this to "GCP" if your Azure credentials are stored in Google Secrets Manager.
 AZURE_AUTH_MODE=default
 
+# Azure resource tag names to include if present, include resourceGroup as a tag name if needed:
+AZURE_RESOURCE_TAG_NAMES=["resourceGroup"] # eg. ["resourceGroup","project","customer"]
+
 # Cache
 
 # Optionally set which cache you are using (defaults to local)
--- a/packages/azure/src/__tests__/ConsumptionManagement.test.ts
+++ b/packages/azure/src/__tests__/ConsumptionManagement.test.ts
@@ -111,6 +111,9 @@ describe('Azure Consumption Management Service', () => {
             serviceName: 'Virtual Machines',
             cost: 5,
             region: 'uksouth',
+            tags: {
+              resourceGroup: 'test-resource-group',
+            },
           },
           {
             kilowattHours: 0.002863085156621724,
@@ -122,6 +125,9 @@ describe('Azure Consumption Management Service', () => {
             serviceName: 'Virtual Machines',
             cost: 10,
             region: 'westeurope',
+            tags: {
+              resourceGroup: 'test-resource-group',
+            },
           },
           {
             kilowattHours: 0.0683736557005491,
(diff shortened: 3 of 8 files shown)


NicolasBissig:
@4upz I have added the azure tagging support as described in #964.

There is a new environment variable `AZURE_RESOURCE_TAG_NAMES` which should be filled with needed tags, and if needed the `resourceGroup`. For example: `AZURE_RESOURCE_TAG_NAMES=["resourceGroup","project","customer"]`. There is currently no way to include all tags, any wishes here?

These tags do not have the same formatting (`user:...` or `azure:...`) as the AWS tags. Is that a problem?

Tests have been updated and expanded to verify inclusion of tags.

It seems that I have to update the templates, but running the script deletes a lot of files that should be needed? Am I doing something wrong?

If something else is different than you expected, please let me know!
Also, when running the App against my Companies azure environment, I discovered three new unknown storage services: `'Data Processing Unit','Pay as you go Warm Storage','Standard Unit'`. Should I add them in this PR as well? All Azure regions were at first marked as unknown, because somehow they are spelled differently when I run CCF, for example: `northeurope` is written as `EU North` and so on. Should I add these new region names?

4upz:
@NicolasBissig thanks so much! I'll give it a review as soon as possible. To answer your questions:

- No need to worry about including all tags. We have this in our roadmap for after tagging support is added for each provider, so it would be beyond the scope of this ticket.
- It's totally fine for them to not match AWS tags. That is just the format in which AWS CUR includes tags in its table. We've included documentation to specify that behavior, and isn't expected for all cloud providers.
- The script shouldn't be deleting files, as it should be adding any changes you made to the api, cli, or client package. However, most of your changes are in the Azure package which does not have a template. So feel free to skip this step.
- That's awesome that you came across unknown services! Yes, please feel free to add them but in a separate PR if you don't mind. You may add the unknown regions to that same PR as well. The regions are due to a weird thing where Azure sometimes returns alternative spellings of regions in its consumption management results. We have a `packages/azure/src/lib/AzureRegions.ts` with all of the regions and known aliases where these missing aliases can be added.

Thanks so much for being thorough and I'll leave any more notes regarding the implementation in my review! However, if you have other questions or if anything is unclear, feel free to let me know 😄

4upz (approved):
Looks great and works perfectly on my end. Thanks again for contributing this @NicolasBissig!

4upz:
I do see some of the tests are failing -- specifically recommendations. However, this could be due to being out of sync with recent changes to the trunk branch. I'll go ahead and merge this in, and will manually fix them if it persists after the merge.

NicolasBissig:
Thank you @4upz !

PR #1027, opened 2023-01-31
Move client emissions call to be global & conditionally show forecast component

### LINKED ISSUE #984: Move Emissions calls to be global instead of page-specific
#### Overview
Our application has two pages, the emissions metrics page and the recommendations page, both of which uses emissions data. As a result, we make two separate requests to our `footprint` endpoint. One request is in the emissions metrics page for the configured start/end date as soon as the app is loaded (since it is the default page). The other request occurs when the recommendations page is loaded, and has a fixed 30-day range from the current day in order to calculate the 30-day forecast in the Forecast component. Fetching fresh emissions can be costly due to cloud provider queries, so to reduce the number of calls we should move the `footprint` request to be global to both pages and to have the recommendations derive its forecast from that fetched data.

#### Out of Scope:

- Checking if individual cloud providers has 30 days available
- Discerning whether data that can be used in Forecast is available in the cache
- Manually estimating 30-day total if emissions data does not contain full 30 days

#### Related issues/Dependencies:

- https://github.com/cloud-carbon-footprint/cloud-carbon-footprint/issues/986
- https://github.com/cloud-carbon-footprint/cloud-carbon-footprint/issues/985
- https://github.com/cloud-carbon-footprint/cloud-carbon-footprint/issues/987

## Acceptance Criteria
- [x] When I first load the CCF client, I only see one emissions request (outside of pagination) being made for all of my emissions data
- [x] When I load the recommendations page while having more than 30 days of data configured/available in my start/end date, I see quicker loading due to my forecasted emissions data already being available
- [x] When I load the recommendations without more than 30 days of data configured/available, I see an error message stating "Not enough data available in request for Forecast. Please adjust your start/end date to include at least 30 days of data"

- [x] Given I have my data grouped by year or quarter, when I navigate to the recommendations page, then I see an error message "In order to see a savings forecast that is relevant to you, please ensure your data is grouped by month, week, or day, and includes data from the past 30 days"
- [x] Given I have my data grouped by month, week or day and I have data for the past 30 days, when I navigate to the recommendations page, then I see a forecast for the last 30 days
- [x] Given I have my data grouped by month, week or day and my data's date range is not in the past 30 days, then I see an error message "In order to see a savings forecast that is relevant to you, please ensure you include data from the past 30 days"
- [x] Given I have my data grouped by month, week or day and my I have data for only some of the past 30 days, then I see a forecast and a warning message "It appears you are missing data for [dates]. Please consider including these days for the most relevant forecast."

## Tech Tasks
- [x] Move the call for the Footprint Service hook to the `App` component
- [x] Pass down the emissions data to the respective pages (components)
- [x] Handle any errors and loading states through prop passing
- [x] Update forecast component logic to check available days based on start/end date, and to display forecast or error message accordingly
- [x] Change affected component tests to accommodate new call structure, handle edge cases 
- [x]  integration-test/recomendations.test.js, EmissionsMetricsPage.test.tsx, RecommendationsPage.test.tsx
- [ ] Update create-app templates
- [ ] If any create-app templates have been updated, make sure to run the [test-create-app script](https://docs.google.com/document/d/1Ei6Cb2DjwEJUNtpULJSxbOuIdAfCARERgE3BSsKWrsc/edit) to ensure it still works properly.

## Tech Notes

- Recommendations component loading state should still be present when grabbing recommendations data and preparing forecast
- Handle logic for changing pages while emissions data is loading. (Should the link be disabled? Should loading be page-specific? Or should there be a global loading screen before pages are loading? Something to discuss)

## Review Tasks
- [ ] Test that data is consistent and present for application after making emissions calls
- [ ] Tests that Forecast error handling is consistent and accurate for each grouping method

ccasher:
## Description of Change

#984 Our application has two pages, the emissions metrics page and the recommendations page, both of which uses emissions data. As a result, we make two separate requests to our footprint endpoint. One request is in the emissions metrics page for the configured start/end date as soon as the app is loaded (since it is the default page). The other request occurs when the recommendations page is loaded, and has a fixed 30-day range from the current day in order to calculate the 30-day forecast in the Forecast component. Fetching fresh emissions can be costly due to cloud provider queries, so to reduce the number of calls we should move the footprint request to be global to both pages and to have the recommendations derive its forecast from that fetched data.

## Checklist

- [ ] yarn test passes
- [ ] yarn lint has been run

## Notes

© 2021 Thoughtworks, Inc.


packages/client/src/utils/helpers/handleDates.ts
```diff
@@ -0,0 +1,79 @@
...
+
+  let endDate: moment.Moment = moment
+    .utc()
+    .subtract(config.MINIMAL_DATE_AGE, 'days')
+  if (config.END_DATE) {
+    endDate = moment.utc(config.END_DATE)
+  }
+
+  let startDate: moment.Moment
+  if (config.PREVIOUS_YEAR_OF_USAGE) {
```
4upz, line 30: Side note: We should add documentation or a note in the `env.template` that makes it clear which date config takes precedence if both exists

packages/client/src/utils/helpers/handleDates.ts
```diff
@@ -0,0 +1,79 @@
...
+  config?: ClientConfig
+}
+
+const groupByAmount = {
+  day: 30,
+  week: 4,
+  month: 1,
+}
+
+export const handleEmissionDateRange = ({
```
4upz: I think this can be renamed to something more clear (i.e. `formatEmissionDateRange` or `getEmissionDateRange`) since it's mostly formatting and retrieving it rather than using it as `handle` implies

packages/client/src/utils/helpers/handleDates.ts
```diff
@@ -0,0 +1,79 @@
...
+
+  return {
+    start: startDate,
+    end: endDate,
+  }
+}
+
+export const checkFootprintDates = ({
+  footprint,
+  groupBy = 'day' as GroupBy,
```
4upz: The default value can probably be simplified to the following: 
```suggestion
  groupBy = GroupyBy.day,
```

However, I'm wondering if we even need a default value or if that's something this function should be concerned with. Isn't the default groupBy set elsewhere?

packages/client/src/utils/helpers/handleDates.ts
```diff
@@ -0,0 +1,79 @@
...
+}): ForecastDetails => {
+  const lastThirtyDays = [...new Array(groupByAmount[groupBy])].map((i, n) =>
+    moment
+      .utc()
+      .startOf(groupBy as moment.unitOfTime.StartOf)
+      .subtract(n, `${groupBy}s` as unitOfTime.DurationConstructor),
+  )
+
+  const footprintDates = []
+  footprint?.data?.map((data) => {
```
4upz: I'm wondering if we need to include all of the footprint data for this, or if we can slice it to the most recent (and presumably last 30) days of data to validate against. That way we're limiting the amount of data we're traversing (saving us time when dealing with a large amount of footprint data)

packages/client/src/utils/helpers/handleDates.ts
```diff
@@ -0,0 +1,79 @@
...
+  const missingDates: Moment[] = lastThirtyDays.filter((a) => {
+    const index = footprintDates.findIndex((date: string) =>
+      moment.utc(date).isSame(a),
+    )
+    return index < 0
+  })
+
+  return {
+    missingDates,
+    groupBy,
```
4upz, line 70: Do we need to return the grouping if the Forecast already has it from the FootprintData?

packages/client/src/utils/helpers/handleDates.test.ts
```diff
@@ -0,0 +1,235 @@
...
+      const footprint = { data, error, loading: false }
+      const { missingDates } = checkFootprintDates({ footprint, groupBy })
+
+      expect(missingDates.length > 0).toBe(true)
+    })
+
+    it('has last 30 day total', () => {
+      const dates = [...new Array(1)].map((i, n) =>
+        moment.utc().startOf(groupBy).subtract(n, `${groupBy}s`).toDate(),
+      )
```
4upz: Should this be a short helper function at the top of the file? Seems like it's being used often

4upz (commented):
Looks pretty good. I marked specific code-related thoughts with my previous comments. I have a few general things that I noticed when testing but can be fixed either before/after merge:
- Page was sometimes not loading all data when enabled for some reason (could be Firefox or cache related though)
- When directly visiting the `/recommendations` page, there are double recommendation requests being made
- Emissions Factors call are still being called every time. This should be moved up as well (or removed if we aren't using it)
- Styling for error message and missing date dropdown needs to be improved (I'll tackle this)
- Dropdown for dates is not centered underneath for larger windows (I'll tackle this)

The most important one is that there were some cases where the forecast warning was inconsistent. For example, I was able to request estimates 3 weeks from today with a grouping of week and still get a forecast (Even though that might not be 30 days?). This could be related to the first point but it may be worth pairing to verifying this before merging

packages/client/src/utils/helpers/handleDates.ts
camcash17: Adding type error here for context, there is probably a way to resolve this but might leave it for a separate fix:
ERROR in ../common/src/index.ts 10:7

Module parse failed: Unexpected token (10:7)
File was processed with these loaders:

```
 * ../../node_modules/@pmmmwh/react-refresh-webpack-plugin/loader/index.js
 * ../../node_modules/source-map-loader/dist/cjs.js
You may need an additional loader to handle the result of these loaders.
| export { default as configLoader, setConfig } from './ConfigLoader'
| export { default as Config } from './Config'
> export type { QUERY_DATE_TYPES, CCFConfig, AWSAccount } from './Config'
| export { GroupBy } from './Config'
| export {
```

camcash17:
> 

Last we spoke, we addressed that the styling concerns can be tackled later, and to ignore the loading data issue for now. I was not able to replicate the emissionsFactor issue, and as far as the recommendations data loading twice, I do see that, and have been unable to fix it so far. It might be worth investigating the useMemo hook in the recommendations data hook file. I am time-boxing this for now and am not sure it needs to be a blocker, but we can continue to debug.

4upz:
Confirming that I'm still seeing:
- Emissions Factors calls are being made twice when first visiting the emissions dashboard (once every time visiting the page, and another after `/footprint` request is made)
image
- Recommendations are still called twice when visiting the page directly (once when page is loaded, second after emissions are called). 
image

Both of these are likely due to the state of the app reloading when estimates are fetched -- thus causing the page (container component) to reload and the page-level request to be made again. Will try to push up a fix with this along with the styles

packages/client/src/utils/helpers/handleDates.ts
4upz: No worries, it's minor so we can leave it as-is

Issue #1022, opened 2023-01-25
Allow Azure Subscriptions to be pulled in chunks instead of all at once


4upz:
## Context
Since we use the Azure Consumption Management API to get usage data, requests for estimates with a large Azure footprint may inevitably encounter rate limits that the API enforces. For these rate limits, we currently apply a retry logic that waits 30 seconds before trying to fetch data from the API again. Rate limit retries can be quite disruptive due to having to periodically wait 60 seconds during every request, and also over-populating the console output for each error encountered. We currently fetch data for all Azure subscriptions all at once, which further increases the chances of running into rate limit errors and needing to retry.

### Related Issue
- #906 
- #1021 

## Tasks

- [x] Implement config option to fetch Azure data in chunks by subscription. This includes an option to enable the chunk method and to specify the chunk size to their liking.
- [x] Implement config option to fetch Azure data synchronously, with the caveat that this will take the longest but will encounter less rate issues.
- [x] Implement config option to fetch Azure data by specific subscriptions/account by providing a list of desired subscriptions. Similar to chunks, but with higher specificity of what exact data to retrieve (instead of all data for the configured account)

## Tech Tasks
- [x] Add appropriate test coverage, including edge cases.
- [x] Run Changesets, bump any packages that have been updated/changed
- [x] Did you make changes to the client, api or cli package? If so, considering updating the create-app templates.
- [x] If any create-app templates have been updated, make sure to run the [test-create-app script](https://docs.google.com/document/d/1Ei6Cb2DjwEJUNtpULJSxbOuIdAfCARERgE3BSsKWrsc/edit) to ensure it still works properly.

PR #1059, opened 2023-02-17
Rework azure rate limit retry logic and add chunking

### LINKED ISSUE #1058: Big Azure estimations never finish due to error in rate limit retry
### Preflight Checklist

- [X] I have read the [Contributing Guidelines](https://github.com/cloud-carbon-footprint/cloud-carbon-footprint/blob/trunk/CONTRIBUTING.md) for this project.
- [ ] I have searched the issue tracker for an issue that matches the one I want to file, without success. **I have found #1022, which could cover this bug as well, but to fix this bug, not all changes in that issue are required. If this is not wanted, feel free to close :)**

### Issue Details

- **Operating System:**
  - Ubuntu 22.04.1 LTS x86_64
- **Browser:**
  - no browser, but CLI 

### Expected Behavior

When I fetch large amounts of Azure data, in my context a year worth of estimates for one Azure subscription using `yarn start-cli --startDate 2022-01-01 --endDate 2023-01-01`, I want to eventually be able to view this data, even if this takes a very long time.

### Actual Behavior

When fetching one month of usage data, rate limit errors occur, but eventually the estimation finishes and I am able to view the data.
When I query a year worth of data, rate limits occur and even after waiting multiple hours in a (CLI has been run in a CI pipeline) the estimations never finish, and no feedback is given, except for rate limit errors.

### To Reproduce

Query a big amount of Azure data at once, sadly I cannot give concrete numbers.

### Screenshots

Log results of querying one month, where after rate limit errors the CLI was able to recover:
image

Log results of querying one year, where after rate limit errors the CLI was never able to recover:
image

### Additional Information

I have taken a look at the code myself, and it seems that the rate limit error handling implementation has a flaw at the moment:

> estimations can only be finished, if the requested time frame is possible to be queried at once without hitting a rate limit.

There exists a [recursive call to the fetch method](https://github.com/cloud-carbon-footprint/cloud-carbon-footprint/blob/trunk/packages/azure/src/lib/ConsumptionManagement.ts#L329), without changing parameters. This causes the rate limited query to be executed without changes, and if the query is too big to finish without hitting rate limits, cause an infinite loop.

I suggest the following changes:

- [ ] implement a maximum retry mechanism, even if it set very high, to avoid infinite loops, and provide feedback to the user
- [ ] provide the possibility to split the queried time range into smaller ranges, and only query these
- [ ] the splitting should be optional, and be configurable as days that should be queried at once

If I misunderstood something in the code, please correct me!

I have implemented these suggestions myself, and I am now able to query one year worth of data.
If these suggestions are something you would want for CCF, I will create a pull request soon, after performing more tests and cleaning up a bit.
At the moment, I am not able to provide all the changes requested in #1022, but maybe this provides a basis for this issue? For my company at least, this is a critical issue.

Hope this helps! 

© 2021 Thoughtworks, Inc.

NicolasBissig:
## Description of Change

fixes #1058 

- [X] added max retries and logging with hint on how to fix
- [X] added support for "day based chunking" of Azure requests to handle large requests, that would otherwise result in an infinite loop
- [X] added config option to turn on chunking and set chunk size in days

## Checklist

- [x] yarn test passes
- [X] yarn lint has been run

## Notes

© 2021 Thoughtworks, Inc.

### DIFF
--- a/packages/azure/src/lib/ConsumptionManagement.ts
+++ b/packages/azure/src/lib/ConsumptionManagement.ts
@@ -311,7 +311,7 @@ export default class ConsumptionManagementService {
   ): Promise<Array<UsageDetailResult>> {
     let currentTry = 0
 
-    const errorMsg = `Azure ConsumptionManagementClient UsageDetailRow paging for time range ${startDate.toISOString()} to ${endDate.getTime()} failed. Reason:`
+    const errorMsg = `Azure ConsumptionManagementClient UsageDetailRow paging for time range ${startDate.toISOString()} to ${endDate.toISOString()} failed. Reason:`
 
     while (currentTry <= maxRetries) {
       currentTry++
--- a/packages/azure/src/lib/ConsumptionManagement.ts
+++ b/packages/azure/src/lib/ConsumptionManagement.ts
@@ -8,6 +8,7 @@ import {
 } from '@azure/arm-consumption'
 import { PagedAsyncIterableIterator } from '@azure/core-paging'
 import {
+  configLoader,
   containsAny,
   convertGigaBytesToTerabyteHours,
   convertTerabytesToGigabytes,
@@ -83,6 +84,7 @@ export default class ConsumptionManagementService {
   private readonly consumptionManagementLogger: Logger
   private readonly consumptionManagementRateLimitRemainingHeader: string
   private readonly consumptionManagementRetryAfterHeader: string
+
   constructor(
     private readonly computeEstimator: ComputeEstimator,
     private readonly ssdStorageEstimator: StorageEstimator,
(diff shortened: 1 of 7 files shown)


packages/azure/src/lib/ConsumptionManagement.ts
```diff
@@ -300,34 +302,98 @@ export default class ConsumptionManagementService {
...
+    startDate: Date,
+    endDate: Date,
+  ): Promise<Array<UsageDetailResult>> {
+    this.consumptionManagementLogger.info(
+      `Getting consumption usage details from ${startDate.toISOString()} to ${endDate.toISOString()}`,
+    )
+    const AZURE = configLoader().AZURE
+
+    // no chunking is wished for
+    if (!AZURE.CONSUMPTION_CHUNKS_DAYS || AZURE.CONSUMPTION_CHUNKS_DAYS === 0) {
```
4upz: 0 is considered a falsy value, so the second condition may not be needed as the first will also return true for a value of 0.
```suggestion
    if (!AZURE.CONSUMPTION_CHUNKS_DAYS) {
```

packages/azure/src/lib/ConsumptionManagement.ts
```diff
@@ -300,34 +302,98 @@ export default class ConsumptionManagementService {
...
+      return await this.queryConsumptionUsageDetails(startDate, endDate)
+    }
+
+    this.consumptionManagementLogger.info(
+      `Time range is going to be requested in chunks of ${AZURE.CONSUMPTION_CHUNKS_DAYS} days`,
```
4upz: Just more concise wording
```suggestion
      `Time range will be requested in chunks of ${AZURE.CONSUMPTION_CHUNKS_DAYS} days`,
```

packages/cli/.env.template
```diff
@@ -60,6 +60,12 @@ AZURE_TENANT_ID=your-azure-tenant-id
 # Optionally set this to "GCP" if your Azure credentials are stored in Google Secrets Manager.
 AZURE_AUTH_MODE=default
 
+# Azure resource tag names to include if present, include resourceGroup as a tag name if needed:
+AZURE_RESOURCE_TAG_NAMES=["resourceGroup"] # eg. ["resourceGroup","project","customer"]
```
4upz, line 64: Nice catch! Thanks for adding this here!

4upz (changes requested):
Hi @NicolasBissig thanks again for implementing this! I've included a few notes from my review. 

The biggest thing that I noticed is that using any form of chunking for requests with multiple subscriptions causes a longer time to execute as well as more frequent rate limits and logging. Sometimes this will also lead to dropped/hanging requests. This is likely due to how we handle multiple requests in [AzureAccount.ts](https://github.com/cloud-carbon-footprint/cloud-carbon-footprint/blob/trunk/packages/azure/src/application/AzureAccount.ts#L82) where we make the request for each subscription in parallel.

I think it's fine to have longer requests when using chunks with multiple accounts, and other fixes and improvements for multiple subscriptions can be incorporated into and implemented with #1022. For the scope of this issue, I think it would be good to add a change in `AzureAccount.ts` where it switches to fetching each subscription sequentially whenever chunking is enabled -- just to ensure that it still plays nice for those with more than one subscription.

Feel free to let me know if you have any questions or thoughts 😄

NicolasBissig:
@4upz thanks for the review, will take a look at it tomorrow!

NicolasBissig:
@4upz I have resolved all mentioned items.

The failing tests also fail on the trunk branch for me, so it seems the failures were not caused by these changes...

Hope all is good now, otherwise let me know!

4upz (approved):
Awesome, the changes look great! Thanks again for working on this @NicolasBissig 

The failing tests were in fact unrelated and were caused by a bug that has been fixed on trunk. So no worries about those. I'll go ahead and merge this in! 😄🚀

PR #1061, opened 2023-02-21
fixed : Support AWS CN

marlkiller:
The current project is not available for AWS CN
I fixed these issue.

### DIFF
--- a/packages/aws/src/application/AWSAccount.ts
+++ b/packages/aws/src/application/AWSAccount.ts
@@ -199,7 +199,7 @@ export default class AWSAccount extends CloudProviderAccount {
       new CloudWatch(options),
       new CloudWatchLogs(options),
       new CostExplorer({
-        region: 'us-east-1',
+        region: configLoader().AWS.IS_AWS_GLOBAL?'us-east-1':'cn-northwest-1',
         credentials: options.credentials,
       }),
       new S3Service(options),
--- a/packages/aws/src/application/AWSCredentialsProvider.ts
+++ b/packages/aws/src/application/AWSCredentialsProvider.ts
@@ -23,9 +23,10 @@ export default class AWSCredentialsProvider {
           configLoader().AWS.authentication.options.proxyRoleName,
         )
       case 'AWS':
+        let partition = configLoader().AWS.IS_AWS_GLOBAL ? `aws`:`aws-cn`
         return new ChainableTemporaryCredentials({
           params: {
-            RoleArn: `arn:aws:iam::${accountId}:role/${
+            RoleArn: `arn:${partition}:iam::${accountId}:role/${
               configLoader().AWS.authentication.options.targetRoleName
             }`,
             RoleSessionName:
--- a/packages/aws/src/application/AWSCredentialsProvider.ts
+++ b/packages/aws/src/application/AWSCredentialsProvider.ts
@@ -23,7 +23,7 @@ export default class AWSCredentialsProvider {
           configLoader().AWS.authentication.options.proxyRoleName,
         )
       case 'AWS':
-        let partition = configLoader().AWS.IS_AWS_GLOBAL ? `aws`:`aws-cn`
+        const partition = configLoader().AWS.IS_AWS_GLOBAL ? `aws`:`aws-cn`
         return new ChainableTemporaryCredentials({
           params: {
             RoleArn: `arn:${partition}:iam::${accountId}:role/${
--- a/packages/common/src/Config.ts
+++ b/packages/common/src/Config.ts
@@ -97,6 +97,11 @@ export type QUERY_DATE_TYPES = {
   [key in GroupBy]: string
 }
 
+const checkAthenaRegionISAWSGlobal = (athena_region: string): boolean => {
+  let AWS_CN_REGIONS = ['cn-north-1','cn-northwest-1']
+  return !AWS_CN_REGIONS.includes(athena_region)
+}
+
 const getAWSAccounts = () => {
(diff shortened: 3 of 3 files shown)


packages/aws/src/application/AWSAccount.ts
```diff
@@ -199,7 +199,7 @@ export default class AWSAccount extends CloudProviderAccount {
       new CloudWatch(options),
       new CloudWatchLogs(options),
       new CostExplorer({
-        region: 'us-east-1',
+        region: configLoader().AWS.IS_AWS_GLOBAL?'us-east-1':'cn-northwest-1',
```
camcash17, line 202: I think rather than a new config called `IS_AWS_GLOBAL`, this could be named `COST_EXPLORER_REGION` where you can specify a preferred region and it would default to `us-east-1`

packages/aws/src/application/AWSCredentialsProvider.ts
```diff
@@ -23,9 +23,10 @@ export default class AWSCredentialsProvider {
           configLoader().AWS.authentication.options.proxyRoleName,
         )
       case 'AWS':
+        let partition = configLoader().AWS.IS_AWS_GLOBAL ? `aws`:`aws-cn`
```
camcash17: Do you know if these are the only two options? If not, I would suggest creating a map based on the region specified in my comment above and using that to set this partition value.

packages/common/src/Config.ts
```diff
@@ -140,6 +141,8 @@ const getConfig = (): CCFConfig => ({
     ATHENA_QUERY_RESULT_LOCATION:
       getEnvVar('AWS_ATHENA_QUERY_RESULT_LOCATION') || '',
     ATHENA_REGION: getEnvVar('AWS_ATHENA_REGION'),
+    // default is aws global
+    IS_AWS_GLOBAL: process.env.IS_AWS_GLOBAL?(process.env.IS_AWS_GLOBAL === 'true'):true,
```
camcash17: See above comment

camcash17 (changes requested):

camcash17:
Hi @marlkiller, thanks so much for submitting this PR! Please take a look at the comments in my review. If these make sense to you, could you also include changes to the documentation? Thanks!

marlkiller:
> Hi @marlkiller, thanks so much for submitting this PR! Please take a look at the comments in my review. If these make sense to you, could you also include changes to the documentation? Thanks!

First of all we are Javascript or Typescript coder. So we just can make some simple changes to integrate the aws-china environment. We are trying to integrate cloud-carbon-footprint with AWS-CHINA.

We answer the pr questions blow: 

 1, So far as we know, AWS only has three partitions, aws, aws-cn and aws-us-gov(we don't work with ever, don't know the region information). and for more information, aws-cn only has two regions. [cn-north-1,cn-northwest-1] 
 
 2, We think we can remove IS_AWS_GLOBAL(environment variable we add) , and then we found that we can get the region information from AWS_ATHENA_REGION(environment variable already exists). As described above so far there is only two regions in aws-cn. We will create a method to determine whether it is aws(global) or aws-cn(china) in arn:


References:
 aws-cn-endpoint : https://docs.amazonaws.cn/en_us/aws/latest/userguide/endpoints-Ningxia.html
 aws-partition : https://docs.aws.amazon.com/general/latest/gr/aws-arns-and-namespaces.html

packages/aws/src/application/AWSAccount.ts
camcash17: I believe based on your latest comment, we should be able to update this to be `configLoader().AWS.ATHENA_REGION` and update it to default to `us-east-1`

packages/aws/src/application/AWSCredentialsProvider.ts
camcash17: this can be a const

packages/common/src/Config.ts
```diff
@@ -140,6 +146,7 @@ const getConfig = (): CCFConfig => ({
     ATHENA_QUERY_RESULT_LOCATION:
       getEnvVar('AWS_ATHENA_QUERY_RESULT_LOCATION') || '',
     ATHENA_REGION: getEnvVar('AWS_ATHENA_REGION'),
```
camcash17, line 148: here is where there can be a default added to be `us-east-1`

packages/common/src/Config.ts
```diff
@@ -96,6 +97,11 @@ export type QUERY_DATE_TYPES = {
   [key in GroupBy]: string
 }
 
+const checkAthenaRegionISAWSGlobal = (athena_region: string): boolean => {
+  let AWS_CN_REGIONS = ['cn-north-1','cn-northwest-1']
```
camcash17: can also be a const

camcash17 (commented):
A few more suggested changes

marlkiller:
I really feel the advice you provided !!
I am a back-end developer and am not proficient in 'Javascript',
Only these simple modifications can be made.

camcash17:
Adding Mark's comment from other branch regarding Athena region:

CostExplorer region is not equal to Athena_region
The region of Cost should be a fixed value

AWS Global : us-east-1
AWS CN : cn-northwest-1

Suppose my athena region is us-east-2, but I can only call the cost api through the us-east-1 region !

Looking forward to your reply.

doc :
https://docs.aws.amazon.com/general/latest/gr/billing.html
https://docs.amazonaws.cn/en_us/aws/latest/userguide/endpoints-Ningxia.html

camcash17:
@marlkiller, I am going to merge this PR as is and provide some further updates directly to the trunk branch following the merge (ie, test fixes, documentation updates, create app/changesets).

Thank you very much for your contribution!

PR #1074, opened 2023-03-09
Removes region from request param

camcash17:
## Description of Change

This PR resolves issue #1070 which is meant to remove 'region' from the estimation request fields. We determined this is not a breaking change and solely used for the unsupported - high fidelity approach.

Including region in the request parameters does not align well with other configurations, and currently there is an option to configure a list of regions. Listing just one region in configuration will have the same effect as previously providing a region in the request parameter.

## Checklist

- [x] yarn test passes
- [x] yarn lint has been run
- [ ] git pre-commit hook is successfully executed. - Precommit hook is currently failing from integration tests
- [ ] relevant documentation is changed or added - There was no documentation prior so I did not update or add anything

## Notes

© 2021 Thoughtworks, Inc.

### DIFF
--- a/.changeset/smooth-pots-begin.md
+++ b/.changeset/smooth-pots-begin.md
@@ -1,0 +1,8 @@
+---
+'@cloud-carbon-footprint/app': minor
+'@cloud-carbon-footprint/cli': minor
+'@cloud-carbon-footprint/client': minor
+'@cloud-carbon-footprint/create-app': patch
+---
+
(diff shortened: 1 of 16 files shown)


4upz (approved):
LGTM! I wasn't able to personally test the Cloud Usage API approach to ensure the flow was still the same. However, the billing approach still works and the logic for both seems to be intact from the code. Gonna merge, but worth keeping an eye on how it progresses.

PR #1087, opened 2023-03-23
feat: add script for mock data

### LINKED ISSUE #532: Improve mock data for demo app
**Context**
To make Cloud Carbon Footprint more quickly understood and more accessible to people who are less technical, we have built out a demo version of the app that uses mock data. Unfortunately there are some issues with this mock data that we would like to clean up. Feel free to do those that you're able to and comment if you can't fix all these issues.

**Improvements**
- [ ] Carbon, energy and cost data on timeline should show more of a correlation
- [ ] When data is grouped by day, the timeline chart should be smoother and more manageable to look at
- [ ] When I filter out data on the recommendations page, the forecast should never be negative (for example, unchecking AWS results in a negative cost for the 30 day forecast - see screenshot)
- [ ] When I view the cost and carbon demo data on the emissions dashboard, the amounts should make sense relative to each other
- [ ] When I view the cost and carbon demo data on the recommendations dashboard, the amounts should make sense relative to each other
- [ ] When I view the carbon and energy demo data on the emissions dashboard, the amounts should make sense relative to each other
- [ ] When I view the carbon and energy demo data on the recommendations dashboard, the amounts should make sense relative to each other

**Out of Scope**
- [ ] Modifying the dates of the mocked data, as that is covered in https://github.com/cloud-carbon-footprint/cloud-carbon-footprint/issues/745

Screen Shot 2021-11-05 at 3 39 32 PM

Samridhi-98:
## Description of Change

Fixes:  #532 

## Checklist

- [x] yarn test passes
- [ ] yarn lint has been run

© 2021 Thoughtworks, Inc.


camcash17:
Hi @Samridhi-98, just checking in! Did you mean to make this a draft PR? Did you want us to start reviewing?

Samridhi-98:
> Hi @Samridhi-98, just checking in! Did you mean to make this a draft PR? Did you want us to start reviewing?

Hi Cam, work is still in progress for this PR.

scripts/index.js
```diff
@@ -0,0 +1,31 @@
+const fs = require('fs')
+const path = require('path')
+//CO2e -> kilowattHours * emissions[region];
```
camcash17: Once this is merge-ready, you can remove unnecessary comments and console logs

scripts/index.js
```diff
@@ -0,0 +1,31 @@
...
+function updateCO2eData() {
```
camcash17: i would suggest making this async and just naming the function `main` to align with our other scripts

scripts/index.js
```diff
@@ -0,0 +1,31 @@
...
+        (err) => {
+            if (err) {
+                console.error(err)
+            }
+        },
+    )
+    console.log(mockData);
+}
+
+updateCO2eData();
```
camcash17: with the above comment about renaming to "main", the calling of the function can be updated to the following:
```
main().catch((error) => {
  console.error(error.stack)
  process.exit(1)
})
```

scripts/index.js
```diff
@@ -0,0 +1,31 @@
...
+//CO2e -> kilowattHours * emissions[region];
+function updateCO2eData() {
+    const data = fs.readFileSync(
+        path.resolve(__dirname, `../packages/client/stub-server/mockData.json`),
+        'utf8',
+    );
+    const mockData = JSON.parse(data);
+
+    mockData.footprint.forEach(({ serviceEstimates }) => {
+        serviceEstimates.forEach((data) => {
```
camcash17: rather than 'data', consider 'serviceEstimate' to avoid confusion with the other data variable in scope defined earlier

scripts/index.js
```diff
@@ -0,0 +1,31 @@
...
+            let emissions = mockData.emissions.find(({ region }) => region === data.region);
+            let co2e = data.kilowattHours * emissions.mtPerKwHour;
```
camcash17: these variables can both be constants

camcash17 (changes requested):

camcash17:
> > Hi @Samridhi-98, just checking in! Did you mean to make this a draft PR? Did you want us to start reviewing?
> 
> Hi Cam, work is still in progress for this PR.

Thanks for working on this! I made a few comments/suggestions in my review and it looks good so far.

A few points I am thinking of:

- Don't worry about cost. That can remain a manually added figure.
- Would you want to consider adding the same update functionality for the Recommendations co2e savings calculation? It would be the same logic multiplying the kwh savings by the emissions factor associated with that region.
- Something important that we want to change is updating the timestamp of the footprint service estimates to remain recent. Would you want to consider enhancing the script to update the time stamps to be the last 15 months from the current date?
- This might be out of scope, but we should consider making the emissions factors in the mockData.json consistent with actual values in our coefficients. For example, [here](https://github.com/cloud-carbon-footprint/cloud-carbon-coefficients/blob/main/data/grid-emissions-factors-aws.csv) is the list for AWS regions. The reason for this is that the mockData is used on our public facing [demo](https://demo.cloudcarbonfootprint.org/) and cloud service providers are expecting it to reflect actual values.

scripts/index.js
camcash17: with this change, I would also request changing the name of the file from index.js to something like 'updateMockData.js'

Samridhi-98:
> 

Thanks for reviewing Cam.
I will make all the changes you suggested.

Samridhi-98:
> > > Hi @Samridhi-98, just checking in! Did you mean to make this a draft PR? Did you want us to start reviewing?
> > 
> > 
> > Hi Cam, work is still in progress for this PR.
> 
> Thanks for working on this! I made a few comments/suggestions in my review and it looks good so far.
> 
> A few points I am thinking of:
> 
> * Don't worry about cost. That can remain a manually added figure.
> * Would you want to consider adding the same update functionality for the Recommendations co2e savings calculation? It would be the same logic multiplying the kwh savings by the emissions factor associated with that region.
> * Something important that we want to change is updating the timestamp of the footprint service estimates to remain recent. Would you want to consider enhancing the script to update the time stamps to be the last 15 months from the current date?
> * This might be out of scope, but we should consider making the emissions factors in the mockData.json consistent with actual values in our coefficients. For example, [here](https://github.com/cloud-carbon-footprint/cloud-carbon-coefficients/blob/main/data/grid-emissions-factors-aws.csv) is the list for AWS regions. The reason for this is that the mockData is used on our public facing [demo](https://demo.cloudcarbonfootprint.org/) and cloud service providers are expecting it to reflect actual values.

Thanks for reviewing Cam.
I will make all the changes you suggested.

camcash17 (approved):

PR #1101, opened 2023-03-29
[#1084] export some objects from package aws

### LINKED ISSUE #1084: Want to have an public method to just calculate the AWS carbon emissions of existing billing data.
Hi,

Very nice project and now I'm using your methodology in my project. But now I meet one problem.

I'd like to calculate the carbon emissions by using our existing AWS billing data, instead of providing the cloud account credential. 
I know that there's a lookup table way to get the result by send the input lookup data (https://www.cloudcarbonfootprint.org/docs/creating-a-lookup-table). But the method is not support to send the usage amount. It needs to multiply the result by myself. 
I attached as below, you can see the usage amount is hard coded as 1.
```
inputData.map((inputDataRow: LookupTableInput) => {
      const costAndUsageReportRow = new CostAndUsageReportsRow(
        null,
        '',
        '',
        inputDataRow.region,
        inputDataRow.serviceName,
        inputDataRow.usageType,
        inputDataRow.usageUnit,
        inputDataRow.vCpus != '' ? parseFloat(inputDataRow.vCpus) : null,
        1,    // hard code as 1 here
        1,
        {},
      )
```

Actually, my AWS billing data is very similar with the "CostAndUsageReportsRow" and I'd like to set up these usage data and build the report data by using my own data model. I realized that, the function "getFootprintEstimateFromUsageRow" in "https://github.com/cloud-carbon-footprint/cloud-carbon-footprint/blob/trunk/packages/aws/src/lib/CostAndUsageReports.ts" is what I want exactly. But the method is declared as private as below.
```
private getFootprintEstimateFromUsageRow(
    costAndUsageReportRow: CostAndUsageReportsRow,
    unknownRows: CostAndUsageReportsRow[],
  )
```

Because I imported the @cloud-carbon-footprint/aws as a 3rd party library in my project, I will not change the source code of it. Can you help change this method to private so I can use it directly?😄

## Possible Implementation

Change the private function "getFootprintEstimateFromUsageRow" to public in "https://github.com/cloud-carbon-footprint/cloud-carbon-footprint/blob/trunk/packages/aws/src/lib/CostAndUsageReports.ts"
```
**public** getFootprintEstimateFromUsageRow(
    costAndUsageReportRow: CostAndUsageReportsRow,
    unknownRows: CostAndUsageReportsRow[],
  ): FootprintEstimate | void
```

HankLeo:
## Description of Change

fix #1084 
export CostAndUsageTypes and getFootprintEstimateFromUsageRow function
As issue #1084 mentioned, in my usage, I imported some of packages as the 3rd party libraries in my project. It's better to use function "getFootprintEstimateFromUsageRow" of "CostAndUsageReports.ts" directly.
Are you willing to export them?

### DIFF
--- a/packages/aws/src/lib/CostAndUsageReports.ts
+++ b/packages/aws/src/lib/CostAndUsageReports.ts
@@ -209,7 +209,7 @@ export default class CostAndUsageReports {
     return result
   }
 
-  private getFootprintEstimateFromUsageRow(
+  getFootprintEstimateFromUsageRow(
     costAndUsageReportRow: CostAndUsageReportsRow,
     unknownRows: CostAndUsageReportsRow[],
   ): FootprintEstimate | void {
--- a/packages/aws/src/lib/index.ts
+++ b/packages/aws/src/lib/index.ts
@@ -12,6 +12,7 @@ export { default as RDSStorage } from './RDSStorage'
 export { default as Lambda } from './Lambda'
 export { ServiceWrapper } from './ServiceWrapper'
 export { default as CostAndUsageReports } from './CostAndUsageReports'
+export * from './CostAndUsageTypes'
 export { AWS_REGIONS } from './AWSRegions'
 export {
   RightsizingRecommendations,


4upz:
Thanks for opening this PR @HankLeo. We're not sure that exporting the function would be a good idea for this use case, as the Lookup Table approach seems to satisfy this. If there is an architectural benefit or larger use case for changing the access to this function, we're happy to explore it. Otherwise, please see my suggestion in #1084

PR #1139, opened 2023-05-12
Update AzureRegions.ts

laellyo:
## Description of Change

Add missing Azure regions based on my logs:

```bash
[AzureRegions] warn: Found unknown azure region 'Europe', please add it to the AzureRegions.ts file and submit a PR, thank you!  
[AzureRegions] warn: Found unknown azure region 'EuropeWest', please add it to the AzureRegions.ts file and submit a PR, thank you! 
[AzureRegions] warn: Found unknown azure region 'qatarcentral', please add it to the AzureRegions.ts file and submit a PR, thank you!  
[AzureRegions] warn: Found unknown azure region 'uaen', please add it to the AzureRegions.ts file and submit a PR, thank you!
```

### DIFF
--- a/packages/azure/src/lib/AzureRegions.ts
+++ b/packages/azure/src/lib/AzureRegions.ts
@@ -71,13 +71,14 @@ export const AZURE_REGIONS = {
   IN_WEST: { name: 'westindia', options: ['westindia'] },
 
   // EU regions
+  EU: { name: 'europe', options: ['Europe'] },
   EU_NORTH: {
     name: 'northeurope',
     options: ['northeurope', 'europenorth', 'NORTHEUROPE', 'EU North'],
   },
   EU_WEST: {
     name: 'westeurope',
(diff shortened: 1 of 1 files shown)


4upz (commented):
Awesome, thanks for adding these @laellyo! I see that a line for a new region was added for just "Europe". This would qualify it as a separate datacenter that we would need to add emissions factors for. 

According to Microsoft's map, they still only have 2 regions in Europe -- Europe North and Europe West. Any insights on if your data that returned this could be actually corresponding to one of those two? If not, we may have to continue to treat that alias as unknown

laellyo:
Hi @4upz,

Thanks for your feedback ! 
You are right, this 'EU' region doesn't seem to be mapped to any existing Azure region.
I have removed it from the PR, the time to have details about those resources.

4upz (approved):
Awesome, looks good! Thanks again for your contribution

Issue #568, opened 2021-11-04
Apply strict null checks across packages


4upz-tw:
#### Overview
To better adhere to Typescript standards and to avoid potential build or runtime errors that users may encounter, it was recommended that we add `"strictNullChecks": true` to our repository's `tsconfig`. In doing so, this introduces a lot of new type errors that relate to how we're doing null checks and handling potentially undefined values that are passed throughout our codebase. This issue focuses on resolving those errors so that we may push this change.

#### Out of Scope:

- Any lint warnings that are not compilation errors _(i.e. "Unexpected Any")_

#### Related issues/Dependencies:

- Refer to the original submitted issue for an example of a gist with type errors: https://github.com/cloud-carbon-footprint/cloud-carbon-footprint/issues/476
- Since this is an ongoing ticket, we encourage any users/contributors who encounter a runtime error caused by this to submit a separate issue so we can immediately focus on addressing it. Refer to this example as one instance in which this can cause a runtime error: https://github.com/cloud-carbon-footprint/cloud-carbon-footprint/issues/475
- You're welcome to do a portion of this work at a time as long as it compiles and runs properly. Please post a comment on what you intend to do and then make a PR with your changes.

## Packages Addressed
- [ ] AWS
- [ ] GCP
- [ ] Azure
- [ ] CLI
- [ ] Core
- [ ] App 

## Tech Tasks
The errors (while plentiful) can span several packages, so we recommend focusing on one at a time. To replicate the issue or to make the errors appear, do the following:
- [ ] Add `"strictNullChecks": true` to the compilerOptions section of the root [`tsconfig.json`](https://github.com/cloud-carbon-footprint/cloud-carbon-footprint/blob/trunk/tsconfig.json) file
- [ ] Run `yarn build` in an individual package and you should be presented with compilation errors.
- [ ] Fix errors by adding additional null checks and adjusting types where necessary

## Pre-Commit Tasks
- [ ] Add appropriate test coverage, including edge cases.
- [ ] Run Changesets, bump any packages that have been updated/changed
- [ ] Did you make changes to the client, api or cli package? If so, considering updating the create-app templates.
- [ ] If any create-app templates have been updated, make sure to run the [test-create-app script](https://docs.google.com/document/d/1Ei6Cb2DjwEJUNtpULJSxbOuIdAfCARERgE3BSsKWrsc/edit) to ensure it still works properly.

## Review Tasks
[...]

PR #1143, opened 2023-05-16
Add Strict Null Checks to ts compiler options for Azure Package

abc516:
## Description of Change

This PR addresses #568 (specifically the Azure tasks). 

## Checklist

- [x] yarn test passes
- [x] yarn lint has been run

## Notes

© 2021 Thoughtworks, Inc.

### DIFF
--- a/packages/azure/jest.config.js
+++ b/packages/azure/jest.config.js
@@ -10,7 +10,7 @@ module.exports = {
   coverageThreshold: {
     global: {
       statements: 95,
-      branches: 91,
+      branches: 81, // Todo increase coverage
       functions: 98,
       lines: 95,
     },
--- a/packages/azure/src/__tests__/AzureAccount.test.ts
+++ b/packages/azure/src/__tests__/AzureAccount.test.ts
@@ -4,6 +4,7 @@
 import {
   EstimationResult,
   GroupBy,
+  Logger,
   LookupTableInput,
   LookupTableOutput,
   RecommendationResult,
@@ -14,6 +15,7 @@ import AzureAccount from '../application/AzureAccount'
 import AzureCredentialsProvider from '../application/AzureCredentialsProvider'
 import ConsumptionManagementService from '../lib/ConsumptionManagement'
 import AdvisorRecommendations from '../lib/AdvisorRecommendations'
+import { AZURE_CLOUD_CONSTANTS } from '../domain'
 
 const mockListSubscriptions = { list: jest.fn(), get: jest.fn() }
 
@@ -380,6 +382,11 @@ describe('Azure Account', () => {
         usageUnit: '10 Hours',
       },
     ]
+    AZURE_CLOUD_CONSTANTS.HDDCOEFFICIENT = undefined
+    AZURE_CLOUD_CONSTANTS.SSDCOEFFICIENT = undefined
+    AZURE_CLOUD_CONSTANTS.NETWORKING_COEFFICIENT = undefined
+    AZURE_CLOUD_CONSTANTS.MEMORY_COEFFICIENT = undefined
+    //AZURE_CLOUD_CONSTANTS.SERVER_EXPECTED_LIFESPAN = undefined
 
     // eslint-disable-next-line @typescript-eslint/no-var-requires
     const AWSAccount = require('../application/AzureAccount').default
@@ -403,11 +410,108 @@ describe('Azure Account', () => {
     const apiError = new Error(errorMessage)
 
     ;(createCredentialsSpy as jest.Mock).mockRejectedValue(apiError)
-
+    AZURE_CLOUD_CONSTANTS.HDDCOEFFICIENT = undefined
+    AZURE_CLOUD_CONSTANTS.SSDCOEFFICIENT = undefined
     const azureAccount = new AzureAccount()
 
     await expect(() => azureAccount.initializeAccount()).rejects.toThrow(
       `Azure initializeAccount failed. Reason: ${errorMessage}`,
     )
+  })
+
+  it('Throws an warning when no subscriptions are available ', async () => {
+    const mockCredentials = {
+      clientId: 'test-client-id',
+      secret: 'test-client-secret',
+      domain: 'test-tenant-id',
+    }
+    ;(createCredentialsSpy as jest.Mock).mockResolvedValue(mockCredentials)
+
+    mockListSubscriptions.list.mockReturnValue([])
+
+    const azureAccount = new AzureAccount()
+    await azureAccount.initializeAccount()
+    await azureAccount.getDataFromAdvisorManagement()
+
+    expect(Logger.prototype.warn).toBeCalled
+  })
+
+  it('Throws an warning when it cant get Advisor recommendation for subscription ', async () => {
+    const mockCredentials = {
+      clientId: 'test-client-id',
+      secret: 'test-client-secret',
+      domain: 'test-tenant-id',
+    }
+    ;(createCredentialsSpy as jest.Mock).mockResolvedValue(mockCredentials)
+
+    mockListSubscriptions.list.mockReturnValue([{ subscriptionId: 'sub-1' }])
+    ;(getRecommendationsSpy as jest.Mock).mockRejectedValue({
+      message: 'error getting recommendation',
+    })
+
+    const azureAccount = new AzureAccount()
+    await azureAccount.initializeAccount()
+    await azureAccount.getDataFromAdvisorManagement()
+
+    expect(Logger.prototype.warn).toBeCalled
+  })
+
+  it('Throws an warning when it cant get data recommendation for subscription ', async () => {
+    const mockCredentials = {
+      clientId: 'test-client-id',
+      secret: 'test-client-secret',
+      domain: 'test-tenant-id',
+    }
+    ;(createCredentialsSpy as jest.Mock).mockResolvedValue(mockCredentials)
+
+    mockListSubscriptions.list.mockReturnValue([{ subscriptionId: 'sub-1' }])
+    ;(getEstimatesSpy as jest.Mock).mockRejectedValue({
+      message: 'error getting recommendation',
+    })
+
+    const azureAccount = new AzureAccount()
+    await azureAccount.initializeAccount()
+    await azureAccount.getDataFromConsumptionManagement(
+      startDate,
+      endDate,
+      grouping,
+    )
+
+    expect(Logger.prototype.warn).toBeCalled
+  })
+
+  it('can use consumption chunk days config', async () => {
+    const mockCredentials = {
+      clientId: 'test-client-id',
+      secret: 'test-client-secret',
+      domain: 'test-tenant-id',
+    }
+    ;(createCredentialsSpy as jest.Mock).mockResolvedValue(mockCredentials)
+
+    mockListSubscriptions.list.mockReturnValue([{ subscriptionId: 'sub-1' }])
+    ;(getEstimatesSpy as jest.Mock).mockRejectedValue({
+      message: 'error getting recommendation',
+    })
+
+    const azureAccount = new AzureAccount()
+    await azureAccount.initializeAccount()
+    jest.mock('@cloud-carbon-footprint/common', () => ({
+      configLoader: jest.fn().mockImplementation(() => {
+        return {
+          AZURE: {
+            CONSUMPTION_CHUNKS_DAYS: 3,
+          },
+        }
+      }),
+    }))
+
+    await azureAccount.getDataFromConsumptionManagement(
+      startDate,
+      endDate,
+      grouping,
+    )
+
+    expect(Logger.prototype.warn).toBeCalled
+    expect(Logger.prototype.info).toBeCalled
   })
 })
--- a/packages/azure/src/__tests__/ConsumptionManagement.test.ts
+++ b/packages/azure/src/__tests__/ConsumptionManagement.test.ts
@@ -428,7 +428,7 @@ describe('Azure Consumption Management Service', () => {
             accountId: subscriptionId,
             accountName: subscriptionName,
             serviceName: 'Bandwidth',
-            cost: 5,
+            cost: 10,
             region: 'uksouth',
             tags: {
               resourceGroup: 'test-resource-group',
@@ -1285,7 +1285,7 @@ describe('Azure Consumption Management Service', () => {
             accountId: subscriptionId,
             accountName: subscriptionName,
             serviceName: 'Bandwidth',
-            cost: 5,
+            cost: 10,
             region: 'uksouth',
             tags: {
               resourceGroup: 'test-resource-group',
--- a/packages/azure/src/application/AzureAccount.ts
+++ b/packages/azure/src/application/AzureAccount.ts
@@ -21,6 +21,7 @@ import {
   UnknownEstimator,
   CloudProviderAccount,
   EmbodiedEmissionsEstimator,
+  EstimateUnknownUsageBy,
 } from '@cloud-carbon-footprint/core'
 import {
   configLoader,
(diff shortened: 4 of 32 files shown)


4upz:
Thanks for opening this @abc516! Will give it a look and post feedback as soon as I can!

packages/azure/src/__tests__/AzureAccount.test.ts
```diff
@@ -249,11 +256,108 @@ describe('Azure Account', () => {
...
+    mockListSubscriptions.list.mockReturnValue([{ subscriptionId: 'sub-1' }])
+    ;(getRecommendationsSpy as jest.Mock).mockRejectedValue({
+      message: 'error getting recommendation',
+    })
+
+    const azureAccount = new AzureAccount()
+    await azureAccount.initializeAccount()
+    await azureAccount.getDataFromAdvisorManagement()
+
+    expect(Logger.prototype.warn).toBeCalled
```
4upz, line 456: I think it would be better if we could modify this tests along with the other ones similar to it by checking that the logger is being called along with the correct message (i.e. `.toBeCalledWith('error getting recommendation')

packages/azure/src/application/AzureAccount.ts
```diff
@@ -88,7 +93,7 @@ export default class AzureAccount extends CloudProviderAccount {
           return await this.getDataForSubscription(
             startDate,
             endDate,
-            subscription.subscriptionId,
+            subscription.subscriptionId || '',
```
4upz: It might be good to reference the variable that you created here
```suggestion
            subscription.subscriptionId || defaultSubscriptionId,
```

packages/azure/src/application/AzureAccount.ts
```diff
@@ -134,13 +139,25 @@ export default class AzureAccount extends CloudProviderAccount {
...
+        AZURE_CLOUD_CONSTANTS.MEMORY_COEFFICIENT || defaultMemoryCoefficient,
+      ),
+      new UnknownEstimator(
+        AZURE_CLOUD_CONSTANTS.ESTIMATE_UNKNOWN_USAGE_BY ||
+          EstimateUnknownUsageBy.COST,
+      ),
       new EmbodiedEmissionsEstimator(
-        AZURE_CLOUD_CONSTANTS.SERVER_EXPECTED_LIFESPAN,
+        AZURE_CLOUD_CONSTANTS.SERVER_EXPECTED_LIFESPAN ||
+          defaultMemoryCoefficient,
```
4upz, line 179: I don't think we need to apply a pipe for default values here, as they are constants and not expected to change values from the ones they're assigned. I would rather recommend taking a deeper look into these classes and seeing if there are functions/constructors that could using better null checking.

packages/azure/src/application/AzureCredentialsProvider.ts
4upz, line 1: This may apply to other files like `AzureAccount.ts` as well, but rather than applying a default ID variable with an empty string, it may be better to instead validate the IDs that are being passed. If they are null, we can log an error message regarding invalid credentials. If true, we follow the usual behavior.

These are cases where we are calling an external API with the potentially null values or using a value for required parameters of another function. It's good to note that sometimes proper null checking can be as simple as a conditional that checks the necessary value exists and handles it accordingly if it doesn't; rather than forcing an unexpected empty value by default. 😃

packages/azure/src/domain/AzureFootprintEstimationConstants.ts
4upz, line 1: Are default values needed for the constants that we know exist? (i.e. `AZURE_CLOUD_CONSTANTS.MEMORY_AVG`) If so, is there a way to do this without adding a new conditional operator (`||`) to the ternary statements?

This is making me wonder if the `AzureAccounts.ts` file would be better off given a single empty object of default values to use if constants are not provided to the instructor, rather than assign these values in each file

packages/azure/src/lib/AdvisorRecommendations.ts
```diff
@@ -84,7 +84,7 @@ export default class AdvisorRecommendations
           const resizingTypes = ['Right-size', 'SkuChange']
           if (
             resizingTypes.includes(
-              recommendation.extendedProperties.recommendationType,
+              recommendation.extendedProperties?.recommendationType || '',
```
4upz: We may not want to assign an empty string here. This may cause the if-statement's conditional to always be true whenever recommendationType is not present since `includes('')` always returns true. We should have an extra conditional here that first checks if recommendationType property exists, then checks to see if the property value is included in resizingTypes.

packages/azure/src/lib/ConsumptionDetailRow.ts
```diff
@@ -101,10 +101,11 @@ export default class ConsumptionDetailRow extends BillingDataRow {
 }
 
 const getConsumptionDetails = (usageDetail: UsageDetailResult) => {
+  const timestamp = new Date()
   const consumptionDetails: Partial<BillingDataRow> = {
     cloudProvider: 'AZURE',
     accountName: usageDetail.subscriptionName,
-    timestamp: new Date(usageDetail.date),
+    timestamp: new Date(usageDetail.date || timestamp),
```
4upz, line 108: You can default this in one line using an empty string instead of timestamp

packages/azure/src/lib/ConsumptionManagement.ts
```diff
@@ -114,8 +118,8 @@ export default class ConsumptionManagementService {
     usageRows
       .filter(
         (consumptionRow) =>
-          new Date(consumptionRow.date) >= startDate &&
-          new Date(consumptionRow.date) <= endDate,
+          new Date(consumptionRow.date || Date.now()) >= startDate &&
+          new Date(consumptionRow.date || Date.now()) <= endDate,
```
4upz: We can probably move the Date parameter to a variable:
`const consumptionRowDate = consumptionRow.date || Date.now()`

4upz: This may require you to convert the arrow function to one that uses curly braces with a return statement

packages/azure/src/lib/ConsumptionManagement.ts
```diff
@@ -80,6 +80,10 @@ import {
   AZURE_EMISSIONS_FACTORS_METRIC_TON_PER_KWH,
 } from '../domain'
 
+const defaultKWHUnit = {
+  total: {},
+}
```
4upz, line 85: I believe this can be applied to providing default constant values at the class or config-level rather than by file

packages/azure/src/lib/ConsumptionManagement.ts
```diff
@@ -570,7 +582,11 @@ export default class ConsumptionManagementService {
       replicationFactor: this.getReplicationFactor(consumptionDetailRow),
     }
 
-    let estimate: FootprintEstimate
+    let estimate: FootprintEstimate = {
+      timestamp: new Date(),
+      kilowattHours: 0,
+      co2e: 0,
+    }
```
4upz, line 591: This is a good default value, however, it will make the condition on line 608 always true (since estimate has a value declared and objects are always truthy). You will need to modify that if statement and others like it to fit this value

packages/azure/src/lib/RightsizingRecommendation.ts
```diff
@@ -35,7 +35,7 @@ export default class RightsizingRecommendation {
       }
       parsedInstanceType = instanceTypeSplit.join(' ')
     }
-    return parsedInstanceType
+    return parsedInstanceType || ''
```
4upz: I'm wondering if we can change the above conditionals to return the result, and have the final return statement return an empty string. That way, we reduce the need for the extra conditional using the `||`

packages/core/src/unknown/UnknownEstimator.ts
```diff
@@ -16,7 +17,14 @@ export enum EstimateUnknownUsageBy {
...
   USAGE_AMOUNT = 'usageAmount',
 }
-
+const kwHdefault = {
+  service: {
+    usageUnit: {
+      cost: 50,
+      kilowattHours: 20,
+    },
+  },
```
4upz, line 26: How did you land on these numbers as default values?

packages/azure/jest.config.js
```diff
@@ -10,7 +10,7 @@ module.exports = {
   coverageThreshold: {
     global: {
       statements: 95,
-      branches: 91,
+      branches: 81, // Todo increase coverage
```
4upz, line 13: I believe this will be able to go back up once we reduce the amount of default conditionals used across each file.

packages/azure/src/__tests__/ConsumptionManagement.test.ts
```diff
@@ -428,7 +428,7 @@ describe('Azure Consumption Management Service', () => {
             accountId: subscriptionId,
             accountName: subscriptionName,
             serviceName: 'Bandwidth',
-            cost: 5,
+            cost: 10,
```
4upz, line 431: Do we know the cause of why these test values needed to be updated? Is it due to the default values added?

4upz (commented):
Thanks for your work on this so far @abc516! These changes are going to be immensely helpful.

I left some feedback across the files as I encountered them. The overall questions/suggestions are:

1. There are a lot of conditionals in the form of `||` statements being added. This seems to be due to covering default values for the AzureConstants that are provided as an argument to the `AzureAccount` class. It may be worth exploring ways to centralize these default values, especially for common objects, at a higher level rather than at a per-file level. In cases like the AzureFootprintEstimationConstants, it may even be worth exploring updating the type or creating a new type extension of an existing type so that we're not doing null checks for values that we know will exist (i.e. `MEMORY_AVG`)
2. For coefficient values, mainly in the core package, we should consider using an average for important coefficients such as PUE instead of a constant value of 1. In these same files, it may be worth exploring how to standardize the types or defaults for cloud constants values that we know we'll need rather than applying empty values
3. Some areas, such as Azure Credentials, should not function with empty or default values since it requires configuration or some information to be given by the user before running the app. For these, consider doing null-checking in the form of an if-exist conditional that logs a warning or throws an error if the required value isn't present. In the case of Azure credentials, this saves us the effort of making an API call with a value such as an empty string that we know will be invalid.

I hope this helps! I think those 3 points will be good areas to make improvements on. But awesome work so far, and feel free to let me know if you have any questions or need clarification regarding any of these points!

packages/azure/src/lib/ConsumptionDetailRow.ts
abc516: Hi @4upz , can you explain a bit what you mean here (maybe w/ a code suggestion)?

packages/core/src/unknown/UnknownEstimator.ts
abc516: @4upz I took them from the awsConstants object in the corresponding test file _packages/core/src/unknown/__tests__/UnknownEstimator.test.ts_ . Do you think this is an acceptable default?

4upz: @abc516 Thanks for the clarification! Those may not be the best values to use as defaults. The values in the test file are usually mock data that is randomly chosen or selected fro ma specific sample group. We may want to rethink a more reasonable value that is retrieved elsewhere or exploring error handling if we can't derive a proper default to use

packages/azure/src/lib/ConsumptionManagement.ts
abc516: Hi @4upz, so should I modify it so line 585 is
```
    const defaultEstimate: FootprintEstimate = {
      timestamp: new Date(),
      kilowattHours: 0,
      co2e: 0,
    }
    let estimate: FootprintEstimate = defaultEstimate;
```
    
and line 608 to be like below, so it's checking defaultEstimate is not the assigned value?

`    if(estimate && estimate !== defaultEstimate)`

packages/azure/src/lib/ConsumptionManagement.ts
abc516: Hi @4upz , where would be a good file to cut this statement to? eg `/packages/core/src/FootprintEstimate.ts`

packages/azure/src/application/AzureAccount.ts
abc516: Hi @4upz: to clarify, for example would it be better to change the MemoryEstimator's constructor to be:
```

  constructor(coefficient?: number) {
   if(coefficient === undefined){
      // throw error;
  }
    this.coefficient = coefficient
  }
```

packages/azure/src/application/AzureCredentialsProvider.ts
abc516: @4upz to clarify, do we want to do something like this for all the configLoader defined authentication parameters in this file?
```

const clientId = configLoader().AZURE?.authentication?.clientId
if(clientId === null){
   // log or throw an error
}
    
```

abc516:
Hi @4upz, I took care of some of the easier-to-implement feedback and asked a couple of follow up questions for the other ones.

4upz:
Thanks @abc516! I'll give them a look!

packages/azure/src/application/AzureAccount.ts
4upz: @abc516 Moreso like this:
```
constructor(coefficient?: number) {
   if(coefficient === undefined){
      // set coefficient to default value
  }
    this.coefficient = coefficient
  }
```

packages/azure/src/application/AzureCredentialsProvider.ts
4upz: @abc516 Yup! That's correct. If null/undefined, throw an error saying credentials are missing before attempting the API call

packages/azure/src/lib/ConsumptionManagement.ts
4upz: @abc2Since these are objects, we can't compare them using `===` as it will always return false since it does not do a deep assertion of the values of each object. We can probably check for the presence or truthy (non-zero) value of a co2e or kilowattHours property by doing:

```
if(estimate.co2e && estimate.kilowattHours)
```

Alternatively, you can use the Rambda library that is in our project to do a deep equals of two values. Assuming Ramdba is imported as `R`:
`if( R.equals(estimate, defaultEstimate)`

4upz:
Sorry for the delay in getting back to you @abc516, had a bit limited availability over the last few weeks. I responded to some of the follow-up questions you have. Feel free to let me know if I missed any -- I can give those changes a review together with your other recent changes once they're pushed.

I also want to really thank you again for your work on this! I know this task is a large one, so we appreciate all of your effort!

packages/azure/src/lib/ConsumptionManagement.ts
```diff
@@ -116,11 +116,13 @@ export default class ConsumptionManagementService {
     const unknownRows: ConsumptionDetailRow[] = []
 
     usageRows
-      .filter(
-        (consumptionRow) =>
-          new Date(consumptionRow.date || Date.now()) >= startDate &&
-          new Date(consumptionRow.date || Date.now()) <= endDate,
-      )
+      .filter((consumptionRow) => {
+        const consumptionRowDate = consumptionRow.date || Date.now()
```
4upz, line 120: Have a default value of the current date (`Date.now()`) may introduce a bug where the row is still included if the user has an endDate that matches the current date. It might be better to have an if-statement that checks if the date property is present, and if it isn't, it returns false (excluding that row)

4upz (changes requested):
Only have a single suggestion for the latest commit. The rest looks good!

abc516:
Hi @4upz . Sorry for the delayed response, I started a new position recently and got very caught up with my work there. I am still very interested in completing this PR. Quick question: would it be possible for you to add the 'Hacktoberfest' tag to the associated GH Issue? Thanks.

4upz:
> Hi @4upz . Sorry for the delayed response, I started a new position recently and got very caught up with my work there. I am still very interested in completing this PR. Quick question: would it be possible for you to add the 'Hacktoberfest' tag to the associated GH Issue? Thanks.

Apologies for the late response, I've been away for a bit as well. No worries and congrats on the new position! If it isn't too late, I went ahead added the Hacktoberfest label. Just let me know when you're ready for me to give this another look

Issue #1138, opened 2023-05-11
Cash amount error in recommendation


bgauchon-reply:
### Preflight Checklist

- [x] I have read the [Contributing Guidelines](https://github.com/cloud-carbon-footprint/cloud-carbon-footprint/blob/trunk/CONTRIBUTING.md) for this project.
- [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

- **Operating System:**
Amazon linux 2
- **Browser:**
Irrelevant

### Expected Behavior

Have right prices in actual behavior

### Actual Behavior

My billing from aws is like 2k/month . recommendations shows last 30 days as 40k/month
I encounter a message missing data from the past 30 days, but data is present in athena and cache file

```
[root@ip-10-0-0-147 cloud-carbon-footprint]# for i in {1..9} ; do echo "2023-04-0$i" ; grep "2023-04-0$i" packages/api/estimates.cache.day.json | wc -l ; done
2023-04-01
1
2023-04-02
1
2023-04-03
1
2023-04-04
1
2023-04-05
1
2023-04-06
1
2023-04-07
1
2023-04-08
1
2023-04-09
1
[root@ip-10-0-0-147 cloud-carbon-footprint]# for i in {0..9} ; do echo "2023-04-1$i" ; grep "2023-04-1$i" packages/api/estimates.cache.day.json | wc -l ; done
2023-04-10
1
2023-04-11
1
2023-04-12
1
2023-04-13
1
2023-04-14
1
2023-04-15
1
2023-04-16
1
2023-04-17
1
2023-04-18
1
2023-04-19
1
[root@ip-10-0-0-147 cloud-carbon-footprint]# for i in {0..9} ; do echo "2023-04-2$i" ; grep "2023-04-2$i" packages/api/estimates.cache.day.json | wc -l ; done
2023-04-20
1
2023-04-21
1
2023-04-22
1
2023-04-23
1
2023-04-24
1
2023-04-25
1
2023-04-26
1
2023-04-27
1
2023-04-28
1
2023-04-29
1
[root@ip-10-0-0-147 cloud-carbon-footprint]# for i in {0..1} ; do echo "2023-04-3$i" ; grep "2023-04-3$i" packages/api/estimates.cache.day.json | wc -l ; done
2023-04-30
1
2023-04-31
0
[...]

PR #1155, opened 2023-05-26
fix: recommendation amount formatting error, recommendation last 30 days error

kevinmonvoisin-reply:
## Description of Change

Issue: https://github.com/cloud-carbon-footprint/cloud-carbon-footprint/issues/1138

In the recommendations tab, there was a couple of errors.
1. Last 30 day total was showing a wrong value. Before it was the sum of every values since the date in the configuration environment file or environment variable. Now it displays the value from last 30 days as expected
2. In the recommendations table, the potential cost saving value was wrong. For instance, if the value was 1,234, it would show 1234. Now, it shows 1,234.

## Checklist

- [x] yarn lint has been run

## Notes

Both @bgauchon-reply and I are working on this PR.

© 2021 Thoughtworks, Inc.

### DIFF
--- a/.changeset/fuzzy-bulldogs-sing.md
+++ b/.changeset/fuzzy-bulldogs-sing.md
@@ -1,0 +1,5 @@
+---
+'@cloud-carbon-footprint/client': patch
+---
+
+Improves recommendations forecast accuracy, and fixes cost formatting errors for local currencies
--- a/packages/client/src/pages/RecommendationsPage/RecommendationsPage.tsx
+++ b/packages/client/src/pages/RecommendationsPage/RecommendationsPage.tsx
@@ -60,7 +60,7 @@ const RecommendationsPage = ({
       <div className={classes.boxContainer}>
         <Grid container spacing={3}>
           <RecommendationsTable
-            emissionsData={recommendations.filteredEmissionsData}
+            emissionsData={slicedFootprint}
             recommendations={recommendations.filteredRecommendationData}
             co2eUnit={co2eUnit}
             forecastDetails={forecastDetails}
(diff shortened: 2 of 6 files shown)


4upz (changes requested):
Thanks for opening this PR and submitting this fix @kevinmonvoisin-reply! 

There are a few changes I'd recommend making before this is merge-ready:

- There are some lint formatting errors in the `transformData.ts` file. I believe these may be highlighted by the Github Actions check. If they're not visible,  I'd recommend running `yarn lint:fix` in the client to fix them.

- For filtering the last 30 days to be calculated, the current function works but it may be doing more work than necessary for larger date ranges. We accomplish a similar task in `RecommendationsPage.tsx` with a function that we use to check if a recent 30-day range exists. I'd recommend looking into reusing the `sliceFootprintDataByLastMonth` function that is imported from the `src/utils/helpers/handleDates.ts` file to directly get the last month of data or to reduce the amount of timestamps to be iterated over.

kevinmonvoisin-reply:
Hi @4upz, sorry for the delay.

I've made the changes in my PR and included the function you pointed out to me.
This time, I also made sure that lint is correct :)

4upz:
No worries and thank you, @kevinmonvoisin-reply! 

The changes look good to me! The only thing left is that there seems to be some tests for `transformData.test.ts` and `RecommendationsServiceHook.test.ts` that are failing and need to be updated to reflect the changes. If you run `yarn test` in the `packages/client` directory, it should show which ones.

Is this something you'd be willing to look into? Happy to take care of it if not.

kevinmonvoisin-reply:
Hi @4upz,
If that's ok for you, can you handle it? Otherwise, I can take a look.

4upz:
@kevinmonvoisin-reply No worries! I went ahead and took care of it. Thanks again for your help on this! I'll merge in the changes

PR #1226, opened 2023-08-24
Update CostAndUsageReports.ts with try/catch error handling to enable unknown data to be gracefully handled

bendemora:
Within the getEstimatesFromInputData() function (line 156 onwards):

1: Wrapped the processing of each row (both from inputData and unknownRows) inside a try-catch block.
2: In case of an error, log the problematic row along with the error message to the console.

## Description of Change

fixes ISSUE 1223

## Checklist

- [X] yarn test passes
- [ ] yarn lint has been run

## Notes

© 2021 Thoughtworks, Inc.

### DIFF
--- a/packages/aws/src/lib/CostAndUsageReports.ts
+++ b/packages/aws/src/lib/CostAndUsageReports.ts
@@ -156,58 +156,67 @@ export default class CostAndUsageReports {
   getEstimatesFromInputData(
     inputData: LookupTableInput[],
   ): LookupTableOutput[] {
-    const result: LookupTableOutput[] = []
-    const unknownRows: CostAndUsageReportsRow[] = []
+    const result: LookupTableOutput[] = [];
+    const unknownRows: CostAndUsageReportsRow[] = [];
 
     inputData.map((inputDataRow: LookupTableInput) => {
-      const costAndUsageReportRow = new CostAndUsageReportsRow(
-        null,
-        '',
-        '',
-        inputDataRow.region,
-        inputDataRow.serviceName,
-        inputDataRow.usageType,
-        inputDataRow.usageUnit,
-        inputDataRow.vCpus != '' ? parseFloat(inputDataRow.vCpus) : null,
-        1,
-        1,
-        {},
-      )
+      try {
+          const costAndUsageReportRow = new CostAndUsageReportsRow(
+            null,
+            '',
+            '',
+            inputDataRow.region,
+            inputDataRow.serviceName,
+            inputDataRow.usageType,
+            inputDataRow.usageUnit,
+            inputDataRow.vCpus != '' ? parseFloat(inputDataRow.vCpus) : null,
+            1,
+            1,
+            {},
+          );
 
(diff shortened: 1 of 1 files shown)


packages/aws/src/lib/CostAndUsageReports.ts
```diff
@@ -156,58 +156,67 @@ export default class CostAndUsageReports {
   getEstimatesFromInputData(
     inputData: LookupTableInput[],
   ): LookupTableOutput[] {
-    const result: LookupTableOutput[] = []
-    const unknownRows: CostAndUsageReportsRow[] = []
+    const result: LookupTableOutput[] = [];
```
camcash17, line 159: I think you can undo these changes where you add the semi-colons. It may not play well with the linter

packages/aws/src/lib/CostAndUsageReports.ts
```diff
@@ -156,58 +156,67 @@ export default class CostAndUsageReports {
...
+            result.push({
+              serviceName: inputDataRow.serviceName,
+              region: inputDataRow.region,
+              usageType: inputDataRow.usageType,
+              vCpus: inputDataRow.vCpus,
+              kilowattHours: footprintEstimate.kilowattHours,
+              co2e: footprintEstimate.co2e,
+            });
+          }
+      } catch (error) {
```
camcash17, line 193: did you confirm that this will in fact reach the catch block? i am not sure that the issue here would actually trigger an error, however, I may be wrong.

packages/aws/src/lib/CostAndUsageReports.ts
```diff
@@ -156,58 +156,67 @@ export default class CostAndUsageReports {
...
+            if (footprintEstimate)
+              result.push({
+                serviceName: inputDataRow.serviceName,
+                region: inputDataRow.region,
+                usageType: inputDataRow.usageType,
+                vCpus: inputDataRow.vCpus,
+                kilowattHours: footprintEstimate.kilowattHours,
+                co2e: footprintEstimate.co2e,
+              });
+        } catch (error) {
```
camcash17, line 211: Same thought as above. I think if the error we are trying to fix is the NaN issue, we may be better to error handle at the root where the arithmetic occurs to aggregate the results. I could also see this potentially being console warnings.

camcash17 (changes requested):
I think this is a great start @bendemora! i have a few questions on the location of the error handling itself and if the issue you are facing would actually lead to the catch blocks.

Also as an FYI - currently this PR is aiming to merge into `cloud-carbon-footprint:changeset-release/trunk`. This should just be `cloud-carbon-footprint:trunk`.

Thank you for taking the time to create this PR!

camcash17:
Hi @bendemora have you had any chance to look at the requested changes / questions?

bendemora:
Hey Cameron,

Really sorry I've not had the ability to put work aside and spend time
trying to code (it's been a really long time since I last wrote any
software, so I had a go at this but it's unsurprising that my approach was
wrong)

Your thoughts on how to improve (or really, introduce) error handling here
would be welcome.

Cheers,

Ben

PR #1277, opened 2023-12-06
Create deploy.yml

Anna-LenaK:
## Description of Change

## Checklist

- [ ] yarn test passes
- [ ] yarn lint has been run

## Notes

© 2021 Thoughtworks, Inc.

### DIFF
--- a/.github/workflows/aws.yml
+++ b/.github/workflows/aws.yml
@@ -1,0 +1,94 @@
+# This workflow will build and push a new container image to Amazon ECR,
+# and then will deploy a new task definition to Amazon ECS, when there is a push to the "trunk" branch.
+#
+# To use this workflow, you will need to complete the following set-up steps:
+#
+# 1. Create an ECR repository to store your images.
+#    For example: `aws ecr create-repository --repository-name my-ecr-repo --region us-east-2`.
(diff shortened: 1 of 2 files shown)


4upz (changes requested):
Hi @Anna-LenaK Thanks for opening this PR! Just a heads up for the future, we'd greatly appreciate it if you could provide a brief description of change to the PR to help the review process and ensure that we're aligned on what the PR is trying to accomplish.

Thanks to the awesome comments in the workflows you provided, I assume these are some templates to help deploy CCF to Amazon ECS. I think this would be a great addition. My only suggestion is that perhaps we should put it in a different directory so that it doesn't conflict or mistakenly get triggered alongside our current workflows. 

Would you mind doing the following:

-  Creating a directory called `.github/workflow-templates` and moving the workflows into that directory. This will add great support for these to be used as [Starter Workflows](https://docs.github.com/en/actions/using-workflows/creating-starter-workflows-for-your-organization)
- Rename the yml files so that they're more descriptive of what they're trying to accomplish. For example, `aws-yml` can be `deploy-to-ecs` and `deploy.yml` can be `deploy-to-s3` (please confirm if that is what `deploy.yml` is trying to do and whether it's related to `aws.yml`)

I think with those changes, we'd be happy to add these 🙂

PR #1314, opened 2024-03-12
Trivial documentation fix: remove 'region' as arg for the CLI

jnioche:
The website still mentions `region` but it has been removed since 1.10.0 via 003c8892


4upz:
@jnioche Thanks for catching this! Definitely makes sense to remove, so I'll merge this in 🚀

4upz (approved):

PR #1321, opened 2024-03-25
Update ADOPTERS.md with Aiven

FloorD:
We're using the Cloud Carbon Footprint tool in our console experience to provide customers with information on their emissions.


ccasher (approved):
Awesome!

Issue #1322, opened 2024-03-27
Use nginxinc/nginx-unprivileged for a more secure and hardened container


lukibahr:
To run the container in a kubernetes environment, client container must be run with root privileges due to nginx. 
In a secured kubernetes environment, this leads to pods blocked by policies due to security restrictions. 

In general, it is highly recommended to run containers with dropping capabilities to avoid security issues.

## Feature Suggestion
To run the container more securily, I would recommend to use nginxinc/nginx-unprivileged. This requires to use the container image in the Dockerfile in packages/client, as well as setting a valid path to the nginx.pid because we overwrite the nginx default config.

**Describe the solution you'd like**

The Dockerfile for packages/client should look like this:

```Dockerfile
FROM node:18-alpine as build
WORKDIR /app
COPY ./package.json ./package.json
COPY ./build /app/build

RUN yarn install --immutable

FROM nginxinc/nginx-unprivileged:1.25 #use nginxinc/nginx-unprivileged
COPY --from=build /app/build /var/www
CMD ["nginx", "-g", "daemon off;"]
```

The nginx.conf should look like the following:

```conf
events {
}

pid /tmp/nginx.pid; # use custom path to pid *(1)

http {
    server {
        listen 8080; # set port 8080 instead of 80
        include /etc/nginx/mime.types;
        root /var/www;
        index index.html index.htm;
        location /api {
            resolver 127.0.0.11 ipv6=off;
            proxy_set_header X-Forwarded-Host $host;
            proxy_set_header X-Forwarded-Server $host;
            proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
            proxy_pass http://api:4000$request_uri;
        }
        location / {
            try_files $uri $uri/ =404;
        }
    }
}
```

**Additional context**
I've built and tested the container locally. The image for testing can be found here: docker.io/lukasbahr/ccfclient/latest-amd64

## Possible Implementation
I'll create a MR for this.

**Further reading***
- *(1) https://github.com/nginxinc/docker-nginx-unprivileged?tab=readme-ov-file#troubleshooting-tips
- https://hub.docker.com/r/nginxinc/nginx-unprivileged

CodeMartian:
Closing this based on the PR being merged in.

lukibahr:
@CodeMartian will the updated image be published in dockerhub? Can't find a newer container image than release-2024-02-11

4upz:
[...]

PR #1323, opened 2024-03-27
#1322 add support for hardened container environments

lukibahr:
## Description of Change

This change hardens the container to use nginxinc/nginx-unprivileged:1.25 container for getting the container up and running in a hardened enviroment, where privileged containers are not allowed or capabilities are restricted.

https://github.com/cloud-carbon-footprint/cloud-carbon-footprint/issues/1322

## Checklist

- [x] yarn test passes
- [x] yarn lint has been run

## Notes

This applies for the frontend (client) application only. Configuration changes for the helm templates are not included.

© 2021 Thoughtworks, Inc.

### DIFF
--- a/docker/nginx.conf
+++ b/docker/nginx.conf
@@ -1,9 +1,10 @@
 events {
 }
+pid /tmp/nginx.pid;
 
 http {
     server {
-        listen 80;
+        listen 8080;
         include /etc/nginx/mime.types;
         root /var/www;
         index index.html index.htm;
--- a/packages/client/Dockerfile
+++ b/packages/client/Dockerfile
@@ -5,6 +5,6 @@ COPY ./build /app/build
 
 RUN yarn install --immutable
 
-FROM nginx:stable-alpine
+FROM nginxinc/nginx-unprivileged:1.25
 COPY --from=build /app/build /var/www
 CMD ["nginx", "-g", "daemon off;"]


olivercodes:
@lukibahr do you think it would be worth exploring the use of the chainguard image? A quick scan with [Trivy](https://www.aquasec.com/products/trivy/) reveals there are a number of vulnerabilities in the nginxinc image, whereas the chainguard image has none at the moment.

caveat: with all vuln scans, it's common to see an alarming list of CVEs, and then review them and realize they are all irrelevant to your own usage (but still requires that time commitment).

Only downside is we would need to contact Chainguard for an OSS license to access specific versions (the `latest` tag of their images is always public). 

https://images.chainguard.dev/directory/image/nginx/versions
https://edu.chainguard.dev/chainguard/chainguard-images/reference/nginx/

lukibahr:
@olivercodes I was considering the chainguard image tho, however i faced the license issue you've mentioned too.  If we're ok with the latest image (what we are doing more ore less, using the nginx-alpine:stable, which is also not directly pinned), I could update the MR. 

For accessing specific versions of chainguard, I do not know the workflow or process how this works with chainguard.

olivercodes:
@lukibahr @camcash17 I opened a dialogue with Chainguard but they don't currently have an OSS policy. Sounds like they may be open to it but it may take a while to work out the details. Probably good to merge this and then come back to it when we have something more solid.

lukibahr:
I second that! In the meantime, I'll test the chainguard image with latest tag.

lukibahr:
@olivercodes  @camcash17  Any news from your side? I suggest, we should merge the change and check chainguard images later.

olivercodes:
+1, good to merge @camcash17

ccasher:
Thanks for the contribution @lukibahr!

lukibahr:
@ccasher will the updated image be published in dockerhub? Can't find a newer container image than release-2024-02-11

ccasher:
@lukibahr, apologies for the delay. doing a release now so should be available for you momentarily!

Issue #839, opened 2022-06-10
co2 emissions estimation incoherent between LUT and UI


updiversity:
### Preflight Checklist

- [ x] I have read the [Contributing Guidelines](https://github.com/cloud-carbon-footprint/cloud-carbon-footprint/blob/trunk/CONTRIBUTING.md) for this project.
- [ x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

### Issue Details

- **Operating System:**
  - macOS 10.13.6 
- **Browser:**
  - Chrome

### Expected Behavior

I would expect for the same time window selected for the billing report (AWS), to obtain the same amount of CO2 total emissions either via using the LUT or via using the UI.

### Actual Behavior

The LUT estimation is way bigger, by a factor of 10k

Any help on ensuring the LUT is properly used ? 

I just multiply the usageAmount by the CO2e factor...

If i look at the code of the following code of the UnknownEstimator

```
export default class UnknownEstimator implements IFootprintEstimator {
  constructor(private estimateKilowattHoursBy: EstimateUnknownUsageBy) {}

  estimate(
    data: UnknownUsage[],
    region: string,
    emissionsFactors: CloudConstantsEmissionsFactors,
    constants: CloudConstants,
  ): FootprintEstimate[] {
    return data.map((data: UnknownUsage) => {
      const estimatedKilowattHours = this.estimateKilowattHours(
        data,
        constants.kilowattHoursByServiceAndUsageUnit,
      )
      const estimatedCo2eEmissions = estimateCo2(
        estimatedKilowattHours,
        region,
        emissionsFactors,
      )
      return {
        timestamp: data.timestamp,
        kilowattHours: estimatedKilowattHours,
        co2e: estimatedCo2eEmissions,
        usesAverageCPUConstant: false,
      }
    })
  }

  private estimateKilowattHours(
    unknownUsage: UnknownUsage,
    kilowattHoursByServiceAndUsageUnit: KilowattHoursByServiceAndUsageUnit,
  ): number {
    const serviceAndUsageUnit =
      kilowattHoursByServiceAndUsageUnit[unknownUsage.service] &&
      kilowattHoursByServiceAndUsageUnit[unknownUsage.service][
        unknownUsage.usageUnit
      ]

    if (serviceAndUsageUnit)
      return (
        (serviceAndUsageUnit.kilowattHours /
          serviceAndUsageUnit[this.estimateKilowattHoursBy]) *
[...]

PR #1358, opened 2024-06-24
Fix create Lookup Table

kakazaro:
## Fix Create Lookup table with incorrect estimation on 1 unit of usage amount
Issue: when I create a Lookup table based on input usage csv, but after multiply value on usage csv with estimation value on Lookup table. It didn't match the value from API `/footprint`

Maybe this is something like the issue on [ISSUE-839](https://github.com/cloud-carbon-footprint/cloud-carbon-footprint/issues/839)

## Checklist

- [x] yarn test passes
- [x] yarn lint has been run

## Notes

© 2021 Thoughtworks, Inc.

### DIFF
--- a/packages/aws/src/lib/CostAndUsageReports.ts
+++ b/packages/aws/src/lib/CostAndUsageReports.ts
@@ -395,7 +395,6 @@ export default class CostAndUsageReports {
       timestamp: rowData.timestamp,
       cost: rowData.cost,
       usageUnit: rowData.usageUnit,
-      usageType: rowData.usageType,
     }
     const unknownConstants: CloudConstants = {
       kilowattHoursByServiceAndUsageUnit:
--- a/packages/aws/src/lib/CostAndUsageReports.ts
+++ b/packages/aws/src/lib/CostAndUsageReports.ts
@@ -194,8 +194,8 @@ export default class CostAndUsageReports {
         inputDataRow.usageType,
         inputDataRow.usageUnit,
         inputDataRow.vCpus != '' ? parseFloat(inputDataRow.vCpus) : null,
-        1,
-        1,
+        parseFloat(inputDataRow.usageAmount || '1') || 1,
+        parseFloat(inputDataRow.cost || '1') || 1,
         {},
       )
       const dateTime = new Date().toISOString()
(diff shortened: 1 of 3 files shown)


4upz:
@kakazaro Thanks for opening this PR and apologies for the time it took to review!

If I am understanding your implementation correctly, it seems you are adding support to pass a custom usage amount and cost as columns in the Lookup Table Input CSV so that it can conditionally calculate estimates according to that usage instead of always giving a unit of 1.

Feel free to correct me if I'm wrong in that understanding. I could see that being a useful feature in a general sense. However, if my understanding is correct and if you are truly seeing discrepancies between the API estimations and the Lookup Table calculations prior to this, could you explain how this approach would fix the issue of those discrepancies? 

Wouldn't passing in a usage amount in your input data, having CCF calculate according to that usage, and then outputting the result accordingly still give you the same calculations as manually multiplying the co2e/energy output by your usage?

jsreqpool:
@4upz we at ReqPOOL worked on this fix together with @kakazaro from TechStackApps since we came across a bug while working with the LUT-approach for one of our customers. The reason for using a LUT is because the customer does not permit their actual cloud usage and cost data to leave their environment, so we need to generate a LUT and then multiply the values in a separate application within their environment.

What we've discovered is that for unknown usage types, the LUT-approach creates different results than the API with direct AWS integration does. This is because kWh for unknown usage types are estimated based on the `cost/usageAmount` ratio, which is information that you don't have (by design) when you create the LUT. This bug #839 is already know for quite a while if I look at its history, but nobody ever got around to fixing it.

The solution we found is that it is enough to pass the correct ratios of `cost/usageAmount` in the input CSV for `create-lookup-table`. so you can work with falsified/randomized data by multiplying all the values created from the [Athena query](https://www.cloudcarbonfootprint.org/docs/creating-a-lookup-table/#aws---athena-query) by a random factor (thus keeping the original ratios while falsifying the true values). @kakazaro's fix allows us to do that by actually taking cost and usageAmount into consideration when generating the LUT. If these values are not present, they default back to 1 as was the case before.

One problem I've found after these changes is that `create-lookup-table` now throws an error when `usageAmount `and `cost ` are not present, you might need to look into that before accepting the PR.

Hope this clears things up and we made a valuable contribution for anybody using the LUT-approach in the future! Let me know if you have any more questions.

